PR #8, opened 2023-08-13
Add unit test for precompile/runtime encoding

ferranbt:
## 📝 Summary

This PR introduces two things:
- A utility to generate a random type from any `abi.Type` specification.
- A unit test that checks the **input encoding** of the precompiles. This unit test does not test the precompiles logic but only they follow the ABI specification as described in the Suave library. 

---


accounts/abi/fuzz_utils.go
```diff
@@ -0,0 +1,121 @@
+package abi
```
ferranbt, line 1: This was implemented in a different file so that we do not have to worry about upstream changes in the future.

core/vm/contracts_suave_test.go
```diff
@@ -0,0 +1,132 @@
...
+		"fetchBids":                 fetchBidsAddress,
+		"newBid":                    newBidAddress,
+		"simulateBundle":            simulateBundleAddress,
+		"submitEthBlockBidToRelay":  submitEthBlockBidToRelayAddress,
+		"buildEthBlock":             buildEthBlockAddress,
+		"confidentialStoreRetrieve": confStoreRetrieveAddress,
+		"confidentialStoreStore":    confStoreStoreAddress,
+	}
+
+	// The objective of the unit test is to make sure that the encoding of the precompile
```
ferranbt, line 85: To highlight this remark. Once code generation is in place. This hack can be removed.

PR #9, opened 2023-08-13
Use foundry to build e2e contracts

ferranbt:
## 📝 Summary

This PR is a proposal to replace the workflow to generate the smart contracts for the end-to-end testing.

Before, both the ABI and the bytecode of the e2e smart contracts had to be compiled and placed manually in the `suave/e2e/contracts.go` file. This approach has a few disadvantages:
- It is not clear how/when the smart contracts were compiled and if they are up to date.
- Some IDEs (my Vscode) have difficulties reading big files that contain smart contract bytecode. In my case, it takes a few seconds to load the file.
- The process is not automated.

This PR introduces `foundry` to build the smart contracts and generate the artifacts and a small helper function to read the artifacts in the end-to-end tests.

By default, `foundry` includes in the artifacts folder the AST of the smart contract which is quite big in size and we do not need it. For this PR, it was around 100k of new lines only because of the AST. In order to remove it, I added a post-process in bash that only keeps `abi` and `bytecode` for the contract.
 
Unlike the previous approach, the artifacts **are not** included in the binary now. I think this approach is okay since we are dealing with end-to-end tests. Otherwise, we could [`embed`](https://pkg.go.dev/embed) the files in the binary too.

As a follow-up PR, I would like to introduce a CI job to make sure that the contracts are up to date.

---


suave/foundry.toml
```diff
@@ -0,0 +1,3 @@
+[profile.default]
```
ferranbt, line 1: TODO: Add a specific Solidity version. 0.8.8?

Ruteri (approved):
We could now remove the standalone bytecodes from e2e/contracts.go in favour of using the compiled artifacts right?

ferranbt:
> We could now remove the standalone bytecodes from e2e/contracts.go in favour of using the compiled artifacts right?

I forgot to do it when I moved the PR to the new repo. Fixed it.

PR #10, opened 2023-08-13
Add code generated stubs for suave lib runtime

ferranbt:
## 📝 Summary

This PR introduces an ad-hoc code generator to produce the "server" interface to receive calls from the Suave library. 

Some notes on the PR:
- It does not remove the `RunOffchain` implementation from the precompiles though it is not necessary anymore. This will be addressed in a follow-up PR.
- It introduces a few hacks (see comments on PR) during code generation that will be addressed in follow-up PRs.
---


suave/gen/main.go
```diff
@@ -0,0 +1,335 @@
...
+
+		if insideTypes {
+			return name
+		}
+		return "types." + name
+	}
+
+	// if the type is an array, return "[]" + the type of the element
+	if typ.T == abi.SliceTy {
+		if typ.Elem.T == abi.TupleTy && slicePointers {
```
ferranbt, line 113: This is a hack because the abi package is not able to pack a slice of pointer structs. It will be fixed in a followup PR.

suave/gen/main.go
```diff
@@ -0,0 +1,335 @@
...
+			str := []string{
+				"var err error",
+			}
+
+			var inputs []string
+
+			// [Step 1]: Unpack if:
+			// 1. If there are more than two input items.
+			// 2. There is one input and the type is not bytes.
+			// 3. The 'extractHint' function is a specific case which takes []input and also unpacks it. TODO: Fix.
```
ferranbt, line 160: Hack to support extractHint. I think it is safe to change extractHint to not unpack here.

core/vm/suave_abis.go
```diff
@@ -273,6 +280,11 @@ var precompilesAbi = mustParseAbi(`[
               "internalType": "address[]",
               "name": "allowedPeekers",
               "type": "address[]"
+            },
+            {
+              "internalType": "string",
```
ferranbt, line 285: Missing version field in Bid struct in the Suave ABI. Not a problem once generated artifacts PR lands.

core/vm/contracts_suave.go
```diff
@@ -155,6 +163,10 @@ func (c *confStoreRetrieve) RunOffchain(backend *SuaveExecutionBackend, input []
 	bidId := unpacked[0].([16]byte)
 	key := unpacked[1].(string)
 
+	return c.runImpl(backend, bidId, key)
+}
+
+func (c *confStoreRetrieve) runImpl(backend *SuaveExecutionBackend, bidId [16]byte, key string) ([]byte, error) {
```
Ruteri, line 169: I wonder if this separation is useful in the general case
Would it make sense to also make the Impl functions return regular types?

core/types/suave_structs.go
```diff
@@ -0,0 +1,32 @@
+// Code generated by suave/gen. DO NOT EDIT.
```
Ruteri, line 1: Maybe it would make sense to separate the ABI generation from the rest of the changes (Impl wrapper)?

core/vm/contracts_suave.go
ferranbt: This separation of `Run` and `runImpl` was only made so that the changes to the repo were as low as possible. My idea is that the implementation of the precompiles itself goes directly to the `backendImpl` eventually where it becomes easier to test.

ferranbt: Also, the `Run` and `runImpl` separation of concerns represent the two layers of the interface between the EVM and the precompiles. `Run` deals with the input/output encoding and `runImpl` with the logic. `Run` is to be deprecated in a followup PR (note it is not in use after the generated code).

The types used in `runImpl` are decided at the SuaveLib library level.

core/types/suave_structs.go
Ruteri: Yeah a separate PR
I see two distinct changes - the autogeneration and the changes to contracts' Run functions (and SuavePrecompiledContractWrapper)

core/vm/contracts_suave.go
Ruteri: Leaving decision up to @dmarzzz and @metachris then

dmarzzz (approved):
LGTM

PR #12, opened 2023-08-14
Deprecate manually defined Suavelib ABI

ferranbt:
## 📝 Summary

This PR is the first cleaning up after #9. Instead of using a manually generated Suave lib ABI description, it relies on the one from the forge-generated artifacts. The artifact is `embed` to ensure it is present in the binary file.

---


dmarzzz (approved):
LSTM!

PR #13, opened 2023-08-14
Add ci validation stage for foundry artifacts

ferranbt:
## 📝 Summary

This PR adds a CI validation to make sure that after changes to the contracts the artifacts are correct. It compiles the smart contracts again and checks if the files in git have changed. To do it I had to:
- Remove the `cache` folder since it keeps a timestamp of the last compilation time. I do not think it is a great loss.
- Recompile the contracts without including the metadata hash which depends on the machine and it does not create reproducible builds.

You can check [this](https://github.com/flashbots/suave-geth/pull/13/commits/caaf2fe3a2ec4077e556d36787700d7b43eb8d4f) commit for an example of the failed CI step.

---


suave/foundry.toml
```diff
@@ -1,3 +1,6 @@
 [profile.default]
 src = 'sol'
-out = 'artifacts'
\ No newline at end of file
+out = 'artifacts'
+bytecode_hash = 'none'
+cache = false
+solc_version = '0.8.19'
```
ferranbt, line 6: New configuration to make the compilation deterministic.

Ruteri (approved):
Let's go

PR #16, opened 2023-08-16
Remove errors join

ferranbt:
## 📝 Summary

`errors.Join` function was included in Go `1.20` but the `go.mod` file uses `1.19` version so it throws an error when running `suavecli`.  

---

### DIFF
--- a/suave/cmd/suavecli/main.go
+++ b/suave/cmd/suavecli/main.go
@@ -3,7 +3,6 @@ package main
 import (
 	"context"
 	"crypto/ecdsa"
-	"errors"
 	"fmt"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


suave/cmd/suavecli/main.go
```diff
@@ -98,16 +97,16 @@ func unwrapPeekerError(rpcErr error) error {
 	}
 	decodedError, err := hexutil.Decode(rpcErr.Error()[20:])
 	if err != nil {
-		return errors.Join(rpcErr, err)
+		return fmt.Errorf("%s: %s", rpcErr, err)
```
Ruteri, line 100: hm, so fmt.Errorf accepts a single %w as the error which we could use
something like `fmt.Errorf("unable to decode non-peeker rpc error %w: %s", rpcErr, err.String())`

Ruteri (approved):

Ruteri:
Alternatively we could move to 1.20 maybe?

suave/cmd/suavecli/main.go
ferranbt: Well, this case both things are errors so it should be maybe `fmt.Errorf("%w, %w)`?

ferranbt:
> Alternatively we could move to 1.20 maybe?

Do not have a strong opinion on this, I prefer to be aligned with `go-ethereum` as much as possible.

PR #17, opened 2023-08-21
Use code generation for Suave lib

ferranbt:
## 📝 Summary

This PR introduces several improvements over the code generation library and the precompile runtime:
- It introduces a `yaml` schema to describe the suave precompiles.
- It generates the Solidity Suave library using the `yaml` schema.
- It merges the work from #10 to use the `yaml` schema too. I decided to put this also in this commit since by using the `yaml` the code generation has been simplified a lot. Before, it was based on string concatenation. Now, it leverages more the Go templates.
- **It introduces breaking compatibilities in several Suave precompiles**. The reasoning is that the changes are not that opinionated and we benefit from having a common standard and simple code generator.
    -  `simulateBundle` packs the input `bytes` with abi. Before, it would send the raw bytes to the Go precompile. This differed from other precompiles of the same form.
    - `submitEthBlockBidToRelay` and `simulateBundle` revert if the precompile fails. Before, they would return the execution result in the function. An alternative to not reverting would be to return a pair (result, ok) from the Go precompile itself.
- It generates the `Suave.json` file with the abi.
- It generates a file with a map of the name and address of each precompile (this could be part of a followup PR if necessary).

### Usage

The generator runs with this command:
```
$ go run suave/gen/main.go [--format] [--write]
Write: ./core/types/suave_structs.go
Write: ./core/vm/contracts_suave_runtime_stub.go
Write: ./suave/artifacts/addresses.go
Write: ./suave/sol/libraries/Suave.sol
Write: ./suave/artifacts/SuaveLib.json
```

- **format**: Format the output.
- **write**: Writes the content to the target files. If not set, outputs the file on the screen.
 
---


suave/sol/libraries/Suave.sol
```diff
@@ -69,105 +64,96 @@ library Suave {
...
     }
-}
 
-function idsEqual(Suave.BidId _l, Suave.BidId _r) pure returns (bool) {
-    bytes memory l = abi.encodePacked(_l);
-    bytes memory r = abi.encodePacked(_r);
-    for (uint i = 0; i < l.length; i++) {
-        if (bytes(l)[i] != r[i]) {
-            return false;
+    function simulateBundle(bytes memory bundleData) internal view returns (uint64) {
```
Ruteri, line 137: In the new paradigm of peekers requiring success, simulateBundle should return (uint64, bool), and the boolean should mark whether simulation succeeded

core/vm/contracts_suave_runtime_stub.go
```diff
@@ -1,190 +1,383 @@
 // Code generated by suave/gen. DO NOT EDIT.
```
Ruteri, line 1: Would probably also make sense to rename the file

suave/sol/libraries/Suave.sol
ferranbt: This will be done in a followup PR.

Ruteri (approved):

PR #25, opened 2023-09-06
Skeleton and basic tests for confidential store engine

Ruteri:
## 📝 Summary

Readme diff: https://github.com/flashbots/suave-geth/pull/25/files?short_path=b335630#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

* Adds a new structure - `ConfidentialStoreEngine` which manages confidential store interfaces (storage and synchronization).
* Adds a simple redis (and miniredis) implementation of confidential data store and synchronization protocol. Each `Store` is now propagated through the transport to other nodes.
* Adds a message signer based on `AccountManager` - `AccountManagerDASigner`. This signer signs messages exchanged through confidential store transport layer.

Changes to the Bid:
* Changes `Id` of the Bid structure to be uuidv5 (SHA-1) of the bid
* Adds `Salt` to the Bid structure, a random uuid
* Adds `AllowedStores` to the Bid structure, an allowlist for which execution nodes are able to modify the bid's data in confidential store 

Other notable changes:
* Refactors how suave stuff is handled in the interpreter - introduces `SuaveContext` in place of the previous `SuaveExecutionBackend`. Removes `SetConfidentialInputs` function.
* Simplifies `ConfidentialStoreBackend`s, moves all of the validation logic to the new `ConfidentialStoreEngine`

## 📚 References

---


suave/backends/redis_pubsub.go
```diff
@@ -0,0 +1,177 @@
...
+	r.client.Close()
+
+	return nil
+}
+
+func (r *RedisPubSub) Subscribe(ctx context.Context) <-chan suave.DAMessage {
+	ch := make(chan suave.DAMessage, 16)
+
+	go func() {
+		for r.ctx.Err() == nil && ctx.Err() == nil {
```
metachris, line 88: if one of these errors is not nil, would it get logged somewhere?

metachris: generally on this function, i think we might need to add reconnect logic in case the redis connection drops 🤔

Ruteri: Will make it an endless loop

cmd/utils/flags.go
```diff
@@ -516,6 +516,18 @@ var (
 		Category: flags.SuaveCategory,
 	}
 
+	SuaveConfidentialPubsubRedisEndpointFlag = &cli.StringFlag{
```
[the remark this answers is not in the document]
Ruteri: They can't be merged - one is the backed for storing data (redis owned by whoever is running the node), and the other is a transport layer - owned by whoever is trusted to run the pubsub. Those are two different redis instances, the transport one is shared among multiple nodes

suave/e2e/redis_transport_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/ethereum/go-ethereum/suave/sdk"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRedisBackends(t *testing.T) {
+	mrStore1 := miniredis.RunT(t)
+	mrStore2 := miniredis.RunT(t)
+	mrPubSub := mrStore1
+
+	fr1 := newFramework(t, WithExecutionNode(), WithRedisStoreBackend(mrStore1.Addr()), WithRedisStoreTransport(mrPubSub.Addr()))
```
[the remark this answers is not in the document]
Ruteri: Hm, yeah that's true

suave/cmd/suavecli/Dockerfile.suavecli
```diff
@@ -0,0 +1,20 @@
+# Build in a stock Go builder container
```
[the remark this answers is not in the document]
Ruteri: I'll remove the whole docker setup for now and we should think about how to best re-introduce it later on

suave/e2e/redis_transport_test.go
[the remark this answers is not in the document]
Ruteri: I changed the store backend to work how you proposed, but the same thing won't quite work for the transport, as it requires a shared redis. I made a helper function that spins up a miniredis instance and returns an opt that is then passed to both frameworks - see if you like it better this way.

suave/backends/redis_pubsub.go
```diff
@@ -0,0 +1,177 @@
...
+	formatRedisBidKey = func(bidId suave.BidId) string {
+		return fmt.Sprintf("bid-%x", bidId)
+	}
+
+	formatRedisBidValueKey = func(bidId suave.BidId, key string) string {
+		return fmt.Sprintf("bid-data-%x-%s", bidId, key) // TODO: should also include the hash of the bid at least
+	}
+)
+
+type RedisPubSub struct {
```
dmarzzz, line 36: I think you also need to include a UUID from the node. Last time I worked with redis pubsub, it would pub it's own message back to itself and you are meant to filter these out with the UUID.

suave/backends/redis_store_backend.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return suave.Bid{}, err
+	}
+
+	return bid, nil
+}
+
+func (r *RedisStoreBackend) Store(bid suave.Bid, caller common.Address, key string, value []byte) (suave.Bid, error) {
+	storeKey := formatRedisBidValueKey(bid.Id, key)
+	err := r.client.Set(r.ctx, storeKey, string(value), time.Second).Err()
```
dmarzzz, line 98: We're not putting a TTL/expiry on this. I think that might be okay for now, I would rather worry about Redis getting too large than it expiring out from under me while testing. But we do need to come to consensus on some TTL before testnet and advertise publicly.

Ruteri: Yeah, I've made it a day for now

suave/backends/redis_pubsub.go
Ruteri: I was thinking to maybe set it to the execution node address, but that would mean nodes can't share it. Uuid is easily spoofed, but we could use both as a way to cut out the messages we send

Ruteri: Check if you like it better

core/vm/suave.go
```diff
@@ -4,40 +4,64 @@ import (
 	"fmt"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
 	suave "github.com/ethereum/go-ethereum/suave/core"
 )
 
+type SuaveContext struct {
```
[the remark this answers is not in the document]
Ruteri: It's the same as BlockContext and TransactionContext in the EVM - it's just the data required for execution, not context in the usual sense

suave/core/types.go
```diff
@@ -2,35 +2,68 @@ package suave
...
+	FetchBidsByProtocolAndBlock(blockNumber uint64, namespace string) []types.Bid
 }
 
 type ConfidentialEthBackend interface {
 	BuildEthBlock(ctx context.Context, args *BuildBlockArgs, txs types.Transactions) (*engine.ExecutionPayloadEnvelope, error)
 	BuildEthBlockFromBundles(ctx context.Context, args *BuildBlockArgs, bundles []types.SBundle) (*engine.ExecutionPayloadEnvelope, error)
 }
 
-type BuildBlockArgs = types.BuildBlockArgs
+type PubSub interface {
```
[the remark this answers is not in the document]
Ruteri: Renamed to `TransportTopic`

suave/backends/miniredis_backends.go
```diff
@@ -0,0 +1,171 @@
...
+)
+
+var (
+	miniredisUpsertTopic = "store:upsert"
+
+	formatMiniredisBidKey = func(bidId suave.BidId) string {
+		return fmt.Sprintf("bid-%x", bidId)
+	}
+
+	formatMiniredisBidValueKey = func(bid suave.Bid, key string) string {
```
metachris, line 22: reuse across backends?

suave/backends/redis_store_backend.go
```diff
@@ -0,0 +1,116 @@
...
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
+	suave "github.com/ethereum/go-ethereum/suave/core"
+	"github.com/go-redis/redis/v8"
+)
+
+var ffStoreTTL = 24 * time.Hour
```
metachris, line 15: `24` could be an env var 🤔

suave/backends/redis_store_backend.go
```diff
@@ -0,0 +1,116 @@
...
+		return err
+	}
+	r.client = client
+
+	return nil
+}
+
+func (r *RedisStoreBackend) Stop() error {
+	if r.cancel == nil || r.client == nil {
+		panic("Stop() called before Start()")
```
metachris, line 51: hmmm perhaps better return an error instead of `panic`? i think panic shouldn't be used in live code if at all possible (and if used the function should be prefixed with `Must`)

metachris: in this case, you could just silently return too

suave/backends/redis_store_backend.go
```diff
@@ -0,0 +1,116 @@
...
+	client, err := connectRedis(r.redisUri)
+	if err != nil {
+		return err
+	}
+	r.client = client
+
+	return nil
+}
+
+func (r *RedisStoreBackend) Stop() error {
```
metachris, line 49: if error result is unused, remove

suave/backends/redis_transport.go
```diff
@@ -0,0 +1,183 @@
...
+	}()
+
+	return ch, cancel
+}
+
+func (r *RedisPubSubTransport) Publish(message suave.DAMessage) {
+	log.Trace("Rebids pubsub: publishing", "message", message)
+	data, err := json.Marshal(message)
+	if err != nil {
+		panic(fmt.Errorf("could not marshal message: %w", err))
```
metachris, line 144: no panic please

suave/backends/redis_transport.go
```diff
@@ -0,0 +1,183 @@
...
+		return err
+	}
+	r.client = client
+
+	return nil
+}
+
+func (r *RedisPubSubTransport) Stop() error {
+	if r.cancel == nil || r.client == nil {
+		panic("Stop() called before Start()")
```
metachris, line 69: no panic please

suave/backends/redis_store_backend.go
Ruteri: It's implementing the `node.Lifecycle` interface - unused in geth as well.
I'll maybe log the error if any appear?

suave/backends/redis_transport.go
Ruteri: For this one I'm kind of on the fence. This should never be the case, and would only appear if someone introduced a bug - in which case a panic is the quickest way to spot the error.
On the other hand it's not critical so it'd be better to not terminate.
I'll return errors

suave/backends/redis_transport.go
Ruteri: On the other hand, under what circumstances would this be the case?
Added a log + early return

suave/backends/redis_store_backend.go
Ruteri: I'd avoid giving a false sense of stability by making this an env var - this is subject to change soon

suave/core/utils.go
```diff
@@ -2,8 +2,33 @@ package suave
...
+	"github.com/google/uuid"
 )
 
+var bidUuidSpace = uuid.UUID{0x42}
+var emptyId [16]byte
+
+func calculateBidId(bid types.Bid) (types.BidId, error) {
+	copy(bid.Id[:], emptyId[:])
+
+	body, err := json.Marshal(bid)
```
ferranbt, line 17: Not a blocker but I would consider doing `keccak(rlp(bid))` instead to have the same standard as the other hashed objects in the client.

Ruteri: I'd defer to a later point, probably once we'll consider having another implementation - right now the id is localized to this function, and I'm not sure if bid will rlp-encode straight away

.dockerignore
```diff
@@ -3,3 +3,5 @@
 build/_workspace
 build/_bin
 tests/testdata
+
+suave/devenv
```
metachris, line 7: this seems wrong to ignore 🤔

Ruteri: This is the dockerignore - we don't want to push the docker-compose.yml as that forces re-building of the geth docker image on every change to the devnet definition - huge time sink.

metachris: lmao thought it was the gitignore. thxxx

suave/core/engine.go
```diff
@@ -0,0 +1,402 @@
...
+
+type ConfidentialStoreEngine struct {
+	ctx    context.Context
+	cancel context.CancelFunc
+
+	backend        ConfidentialStoreBackend
+	transportTopic StoreTransportTopic
+	mempool        MempoolBackend
+
+	daSigner    DASigner
```
dmarzzz, line 24: for clarity can we add a comment that "DA = data availability"

dmarzzz (approved):
LSTM

README.md
```diff
@@ -298,26 +303,15 @@ type ConfidentialEthBackend interface {
...
 
-The current, and certainly not final, implementation of the Confidential Store is managed by the `LocalConfidentialStore` struct. It provides thread-safe access to the bids' confidential data. The `LocalConfidentialStore` struct is composed of a mutex lock and a map of bid data, `ACData`, indexed by a `BidId`.
+The current, and certainly not final, implementation of the Confidential Store is managed by the `ConfidentialStoreEngine`. The engine consists of a storage backend, which holds the raw data, and a transport topic, which relays synchronization messages between nodes.  
+We provide two storage backends to the confidential store engine: the `LocalConfidentialStore`, storing data in memory in a simple dictionary, and `RedisStoreBackend`, storing data in redis. To enable redis as the storage backed, pass redis endpoint via `--suave.confidential.redis-store-endpoint`.
```
dmarzzz, line 307: perhaps just add a note that Redis is only intended for first testnet and as an upgrade path for a future p2p based protocol

ferranbt (approved):

PR #27, opened 2023-09-06
Add 'make suave'

andytudhope:
## 📝 Summary

A superficial bit of fun. Are deeper changes required?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,10 +8,11 @@ GOBIN = ./build/bin
 GO ?= latest
 GORUN = env GO111MODULE=on go run
 
-suave:
+suave: geth
 	$(GORUN) build/ci.go install ./cmd/geth
+	cd $(GOBIN) && ln -s geth suave
 	@echo "Done building."
-	@echo "Run \"$(GOBIN)/geth\" to launch suave."
+	@echo "Run \"$(GOBIN)/suave\" to launch suave."
 
 all:
 	$(GORUN) build/ci.go install
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ suave: geth
 	$(GORUN) build/ci.go install ./cmd/geth
 	cd $(GOBIN) && ln -s geth suave
 	@echo "Done building."
-	@echo "Run \"$(GOBIN)/suave\" to launch suave."
+	@echo "Run \"$(GOBIN)/suave\" to launch SUAVE."
 
 all:
 	$(GORUN) build/ci.go install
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -8,10 +8,10 @@ GOBIN = ./build/bin
 GO ?= latest
 GORUN = env GO111MODULE=on go run
 
-geth:
+suave:
 	$(GORUN) build/ci.go install ./cmd/geth
 	@echo "Done building."
-	@echo "Run \"$(GOBIN)/geth\" to launch geth."
+	@echo "Run \"$(GOBIN)/geth\" to launch suave."
```
dmarzzz, line 14: shouldn't this be Run \"$(GOBIN)/suave\ to launch suave"? Which additionally requires modifying to make sure the build outputs to that directory. Also we typically stylize SUAVE but not sure thats necessary here just a thought

dmarzzz (changes requested):
If you search "make geth" in the repo you will find 2 yml files for github workflows to change + the ReadMe.

changing the actually cmd/geth folder would be a little much at this point in time imo but Im fine with this at a high level for aesthetic purposes

Makefile
```diff
@@ -13,6 +13,12 @@ geth:
 	@echo "Done building."
 	@echo "Run \"$(GOBIN)/geth\" to launch geth."
 
+suave:
+	$(GORUN) build/ci.go install ./cmd/geth
+	cd $(GOBIN) && ln -s geth suave
```
metachris, line 18:
```suggestion
	cd $(GOBIN) && rm -f suave && ln -s geth suave
```

need to remove first, otherwise getting this error on repeated executions:

```
cd ./build/bin && ln -s geth suave
ln: failed to create symbolic link 'suave': File exists
make: *** [suave] Error 1
```

Makefile
```diff
@@ -13,6 +13,12 @@ geth:
 	@echo "Done building."
 	@echo "Run \"$(GOBIN)/geth\" to launch geth."
 
+suave:
```
metachris, line 16: you need to also add `suave` to `.PHONY:` in line 5

Makefile
andytudhope: Thanks for catching - have attended to both comments here :pray:

metachris (approved):
nice, thanks!

ferranbt (changes requested):
You have to change also the Dockerfile to use the new suave binary instead of geth

ferranbt (changes requested):
My bad, the `Dockerfile` uses a different build process than the makefile to build the client. No changes are required. You can revert https://github.com/flashbots/suave-geth/pull/27/commits/b298b859a13b08dc976e9b99ce9cbbabfe1acd6d.

metachris (approved):

PR #35, opened 2023-09-18
Use a custom genesis.json file during chain start

ferranbt:
## 📝 Summary

In Geth, there are two ways to select the chain to run:
- There are some hardcoded genesis definitions for the popular chains that are selected with flags (i.e. `--mainnet`, `--sepolia`, etc...).
- You can use a custom `genesis.json` file. This option requires two steps. First, using the `init` command to process the `genesis.json` and execute the genesis block and then the `geth` command to start the chain.

This PR simplifies the second option and enables running a custom `genesis.json` in a single `geth` command:
```
$ geth --chain genesis.json
```

The benefits of this PR are:
- It is easier to deploy custom private PoA networks by running only one command.
- As a follow-up PR we can expand what the `genesis.json` covers and empower it more, for example, add the bootstrap nodes of the chain or the initial set of Clique signers. (this is already on the works).
- It is easier to manage a list of `genesis.json` files than a set of hardcoded Go specifications.

## 📚 References

---


cmd/utils/flags.go
```diff
@@ -163,6 +165,11 @@ var (
 		Usage:    "suave network: pre-configured proof-of-authority suave test network",
 		Category: flags.EthCategory,
 	}
+	ChainFlag = &cli.StringFlag{
+		Name:     "chain",
```
Ruteri, line 169: Maybe `CustomChain`? Or even `CustomGenesis`?

cmd/utils/flags.go
```diff
@@ -1891,6 +1899,27 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 		if !ctx.IsSet(MinerGasPriceFlag.Name) {
 			cfg.Miner.GasPrice = big.NewInt(1)
 		}
+	case ctx.IsSet(ChainFlag.Name):
+		// initialize the chain from a genesis file
+		genesis, genesisHash, err := initCustomGenesis(ctx, stack, ctx.String(ChainFlag.Name))
+		if err != nil {
+			Fatalf("Failed to initialize genesis from file: %v", err)
+		}
+		cfg.Genesis = genesis
```
Ruteri, line 1908: Do we have to do anything with the data dir?
We could, say, put `if ctx.IsSet(ChainFlag.Name) && cfg.DataDir == Default { cfg.DataDir = '....' } in `SetDataDir`

cmd/utils/flags.go
```diff
@@ -1891,6 +1899,27 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 		if !ctx.IsSet(MinerGasPriceFlag.Name) {
 			cfg.Miner.GasPrice = big.NewInt(1)
 		}
+	case ctx.IsSet(ChainFlag.Name):
```
Ruteri, line 1902: We should probably add this flag to `CheckExclusive` at the top of `SetEthConfig`

Ruteri (approved):

cmd/utils/flags.go
ferranbt: Not sure I understand the objective of this.

Ruteri: It's so that a custom genesis won't overwrite the mainnet default directory

ferranbt: Ok, that is not a problem because for specific chains a sub-directory is created.

PR #49, opened 2023-09-29
Merge confidential store and mempool

ferranbt:
## 📝 Summary

This PR merges the confidential store and the mempool (a local indexer on top of the store) in the same interface/building block. Both modules work together and they even cross-reference one another.

## 📚 References

---

### DIFF
--- a/core/vm/suave.go
+++ b/core/vm/suave.go
@@ -11,6 +11,18 @@ import (
 	suave "github.com/ethereum/go-ethereum/suave/core"
 )
 
+// ConfidentialStore represents the API for the confidential store
+// required by Suave runtime.
+type ConfidentialStore interface {
+	Start() error
+	Stop() error
+	InitializeBid(bid types.Bid, creationTx *types.Transaction) (types.Bid, error)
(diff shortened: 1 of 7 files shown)


core/vm/suave.go
```diff
@@ -20,30 +30,10 @@ type SuaveContext struct {
 }
 
 type SuaveExecutionBackend struct {
-	ConfidentialStoreEngine *suave.ConfidentialStoreEngine
-	MempoolBackend          suave.MempoolBackend
+	ConfidentialStoreEngine ConfidentialStore
 	ConfidentialEthBackend  suave.ConfidentialEthBackend
 }
 
-func (b *SuaveExecutionBackend) Start() error {
```
ferranbt, line 28: After the changes, I think the `ConfidentialStore` is a first-class citizen of the system and the initialization/lifecycle should be done as part of the whole `Ethereum` protocol initialization (piggybacking the `stack.RegisterLifecycle` utility) rather than doing it in the `SuaveExecutionBackend` that right now is only a wrapper around the `Suave API` for the precompiles.

Ruteri (approved):

core/vm/suave.go
Ruteri: I'd say that it's still the engine that should manage the store

PR #50, opened 2023-09-29
Simplify dastore backend

ferranbt:
## 📝 Summary

This PR simplifies the Confidential Store implementations to a single Redis implementation. The other two (removed in this PR) are versions of that Redis implementation.

## 📚 References

---


suave/backends/miniredis_backends.go
```diff
@@ -1,160 +0,0 @@
...
-	if err != nil {
-		return err
-	}
-
-	r.client.Set(key, string(data))
-
-	return nil
-}
-
-func (r *MiniredisBackend) FetchEngineBidById(bidId suave.BidId) (suave.Bid, error) {
```
ferranbt, line 125: These functions were copy/paste from the `Redis` backend.

suave/backends/redis_store_backend.go
```diff
@@ -29,6 +30,11 @@ type RedisStoreBackend struct {
 	cancel   context.CancelFunc
 	redisUri string
 	client   *redis.Client
+	local    *miniredis.Miniredis
+}
+
+func NewLocalConfidentialStore() *RedisStoreBackend {
```
ferranbt, line 36: We can use the `miniredis` version of the `RedisStoreBackend` that works in memory as the local store.

suave/backends/redis_store_backend.go
```diff
@@ -39,6 +45,16 @@ func NewRedisStoreBackend(redisUri string) *RedisStoreBackend {
 }
 
 func (r *RedisStoreBackend) Start() error {
+	if r.redisUri == "" {
+		// create a mini-redis instance
```
ferranbt, line 49: It defaults to local `miniredis` if no endpoint is specified. Then, the rest of the store works for both remote/local Redis.

Ruteri (approved):

suave/backends/cdas.go
```diff
@@ -1,80 +0,0 @@
...
-	"errors"
-	"fmt"
-	"sync"
-
-	"github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/log"
-	suave "github.com/ethereum/go-ethereum/suave/core"
-)
-
-type LocalConfidentialStore struct {
```
Ruteri, line 13: I'd maybe leave local store in since it's easy to debug with

PR #57, opened 2023-10-02
Transactional store

Ruteri:
## 📝 Summary

* Adds `TransactionalStore`- a structure that batches updates to the confidential engine
* Rewrites `ConfidentialStoreEngine` to account for batched updates - `InitializeBid` now does not store anything on the backend, and `Store` function is now absent. Adds `Finalize` function to the engine, which writes a batch of updates to the store backend
* Unifies MEVM execution in eth api behind runMEVM function
* Removes transactions from `ConfidentialStore` interface, passing transaction around is now managed by `TransactionalStore`
* Splits `EthAPIBackend.GetEVM` into two - suave-only `GetMEVM` and non-suave only `GetEVM`
* `EthAPIBackend.GetMEVM` now also returns a finalization function which causes the batched writes to confidential store to be applied
* Minor QoL improvements in testing

## 📚 References

---

### DIFF
--- a/suave/backends/transactional_store_test.go
+++ b/suave/backends/transactional_store_test.go
@@ -1,0 +1,82 @@
+package backends
+
+import (
+	"math/big"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+	suave "github.com/ethereum/go-ethereum/suave/core"
+	"github.com/stretchr/testify/require"
+)
+
+func TestTransactionalStore(t *testing.T) {
+	engine := suave.NewConfidentialStoreEngine(NewLocalConfidentialStore(), suave.MockTransport{}, suave.MockSigner{}, suave.MockChainSigner{})
+
+	testKey, _ := crypto.HexToECDSA("b71c71a67e1177ad4e901695e1b4b9ee17ae16c6668d313eac2f96dbcda3f291")
+	dummyCreationTx, err := types.SignTx(types.NewTx(&types.ConfidentialComputeRequest{
+		ExecutionNode: common.Address{0x42},
+		Wrapped:       *types.NewTransaction(0, common.Address{}, big.NewInt(0), 0, big.NewInt(0), nil),
+	}), types.NewSuaveSigner(new(big.Int)), testKey)
+	require.NoError(t, err)
+
+	tstore := engine.NewTransactionalStore(dummyCreationTx)
+
+	testBid, err := tstore.InitializeBid(types.Bid{
+		Salt:                suave.RandomBidId(),
+		DecryptionCondition: 46,
+		AllowedStores:       []common.Address{{0x42}},
+		AllowedPeekers:      []common.Address{{0x43}},
+		Version:             "v0-test",
+	})
+	require.NoError(t, err)
+
+	_, err = tstore.Store(testBid.Id, testBid.AllowedPeekers[0], "xx", []byte{0x44})
+	require.NoError(t, err)
+
+	tfetchedBid, err := tstore.FetchBidById(testBid.Id)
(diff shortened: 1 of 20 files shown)


suave/core/engine.go
```diff
@@ -185,82 +163,74 @@ func (e *ConfidentialStoreEngine) FetchBidsByProtocolAndBlock(blockNumber uint64
...
-		Value:     value,
-		StoreUUID: e.storeUUID,
-	}
+	return e.backend.Retrieve(bid, caller, key)
+}
 
-	msgBytes, err := SerializeMessageForSigning(msg)
-	if err != nil {
-		return Bid{}, fmt.Errorf("confidential engine: could not hash message for signing: %w", err)
+func (e *ConfidentialStoreEngine) Finalize(tx *types.Transaction, newBids map[BidId]Bid, stores []StoreWrite) error {
```
ferranbt, line 179: Would it be possible to only pass the execution node instead of the whole `types.Transaction`?

suave/backends/transactional_store_test.go
```diff
@@ -0,0 +1,82 @@
+package backends
```
ferranbt, line 1: Is this file here because of cycle dependency?

suave/core/engine.go
ferranbt, line 179: Out of the scope of the review, but would this be possible?
```suggestion
type ConfStoreUpdate struct {
   Bid Bid
   Writes []ConfStoreWrite
}
type ConfStoreWrite struct {
   Key []byte
   ...
}
func (e *ConfidentialStoreEngine) Finalize(tx *types.Transaction, updates []ConfStoreUpdate) error {
```

suave/core/engine.go
```diff
@@ -284,71 +254,106 @@ func (e *ConfidentialStoreEngine) NewMessage(message DAMessage) error {
...
 	}
 
 	return nil
 }
 
-func SerializeBidForSigning(bid Bid) ([]byte, error) {
-	bid.Signature = []byte{}
-
-	bidBytes, err := json.Marshal(bid)
+func SerializeBidForSigning(bid *Bid) ([]byte, error) {
```
ferranbt, line 333:
```suggestion
func (bid *Bid) SerializeForSigning() ([]byte, error) {
```

suave/core/engine.go
```diff
@@ -284,71 +254,106 @@ func (e *ConfidentialStoreEngine) NewMessage(message DAMessage) error {
...
 	}
 
 	return []byte(fmt.Sprintf("\x19Suave Signed Message:\n%d%s", len(bidBytes), string(bidBytes))), nil
 }
 
-func SerializeMessageForSigning(message DAMessage) ([]byte, error) {
-	message.Signature = []byte{}
-
-	msgBytes, err := json.Marshal(message)
+func SerializeMessageForSigning(message *DAMessage) ([]byte, error) {
```
ferranbt, line 350:
```suggestion
func (message *DAMessage) SerializeForSigning() ([]byte, error) {
```
We could do a common interface for objects that get signed.

ferranbt (approved):

suave/core/engine.go
Ruteri: Sadly no, the whole transaction is needed

suave/backends/transactional_store_test.go
Ruteri: Kind of, it relies on the local store

suave/core/engine.go
Ruteri: Hm, no - the individual writes are not tied to any one bid. They each have their own bid

suave/core/engine.go
Ruteri: Dunno, I'd much rather keep the bid a data-only structure

suave/core/engine.go
Ruteri: We could, on the other hand I don't think this is too much

PR #66, opened 2023-10-04
Add new call Precompile

ferranbt:
## 📝 Summary

This PR creates a new precompile to query another smart contract from the execution node.

## 📚 References

---


suave/sol/libraries/Suave.sol
```diff
@@ -105,6 +107,15 @@ library Suave {
         }
     }
 
+    function ethcall(address contractAddr, bytes memory input1) internal view returns (bytes memory) {
```
Ruteri, line 110: Hm, how would we handle errors? Do we expect to return the encoded error in bytes here and is that the case?

Ruteri (approved):

PR #76, opened 2023-10-26
Add Forge foundry

ferranbt:
## 📝 Summary

This PR adds some basic support to debug Suave apps in Solidity with Foundry. It introduces a generated Suave contract that implements the precompiles by making `vm.ffi` calls to an external suave command to execute the precompiles.

## 📚 References

---


internal/ethapi/api.go
```diff
@@ -1944,7 +1944,7 @@ func (s *TransactionAPI) SendRawTransaction(ctx context.Context, input hexutil.B
 }
 
 // TODO: should be its own api
-func runMEVM(ctx context.Context, b Backend, state *state.StateDB, header *types.Header, tx *types.Transaction, msg *core.Message) (*types.Transaction, *core.ExecutionResult, func() error, error) {
+func runMEVM(ctx context.Context, b Backend, state *state.StateDB, header *types.Header, tx *types.Transaction, msg *core.Message, isCall bool) (*types.Transaction, *core.ExecutionResult, func() error, error) {
```
Ruteri, line 1947: Nit: would it be clearer to pass in the `vm.Config`?

ferranbt: For now, there is no other field in `vm.Config` that we want to configure. I would do it if that is the case.

core/vm/contracts_suave.go
```diff
@@ -123,6 +124,8 @@ func (c *confStoreStore) runImpl(suaveContext *SuaveContext, bidId suave.BidId,
 		return suave.ErrBidNotFound
 	}
 
+	log.Info("confStoreStore", "bidId", bidId, "key", key)
```
Ruteri, line 127: probably not needed

Ruteri (approved):

core/vm/contracts_suave.go
ferranbt: We do have the same log for retrieve, do you want to remove that one too?

Ruteri: I think so

Ruteri (approved):

PR #80, opened 2023-10-27
Adds pebble db store backend

Ruteri:
## 📝 Summary

* Adds a new Pebble confidential store backend

## 📚 References

---


suave/e2e/pebble_test.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+func WithPebbleDB(t *testing.T) frameworkOpt {
+	tmpDir := t.TempDir()
+	return func(c *frameworkConfig) {
+		c.suaveConfig.PebbleDbPath = tmpDir
+	}
+}
+
+func TestPebbleStore(t *testing.T) {
```
ferranbt, line 20: There is a [generic test for the backends](https://github.com/flashbots/suave-geth/blob/main/suave/cstore/redis_store_backend_test.go#L7) so that we do not have to rely on `e2e` to cover them. Can you use that instead?

ferranbt (approved):

PR #88, opened 2023-10-31
Precompile dispatch table

ferranbt:
## 📝 Summary

This PR introduces a concept developed as part of #70, the `dispatcher`. The `dispatcher` is an object that stores all the list of MEVM precompiles and can invoke each one of them on runtime without needing the generated code for encoding/decoding.

### Writing a new precompile

Now, writing a precompile is two steps.

First, write a struct that fulfills the `SuavePrecompile` interface:

```
type addPrecompile struct {}

func (a *addPrecompile) Address() common.Address {
   return common.Address("0x00...")
}

func (a *addPrecompile) Name() string {
   return "add"
}

func (a *addPrecompile) RequiredGas(input []byte) uint64 {
  return 100
}

func (a *addPrecompile) Do(ctx *SuaveContext, a, b uint64) (uint64, error) {
   return a + b
}
```

Methods:
- `Address`: Returns the address assigned to this precompile.
- `Name`: Returns the name of the precompile. This will be the name used on the `Suave.sol` library.
- `RequiredGas`: The gas required by the precompile.
- `Do`: A function with the **typed** inputs/outputs for the precompile. It must have as a first argument `SuaveContext` and as last output an `error`.

Second, the precompile gets registered in the dispatcher:

```
func init() {
   dispatcher.MustRegister(&addPrecompile{})
}
```

### Suave library code generation

With the introduction of the dynamic runtime, the Go code generation is not required anymore.

Besides, now, it is the Go precompiles and the dispatcher the new source of truth about the precompiles. Thus, the yaml specification is not required anymore.

The code generation is still used to generate `Suave.sol` but it uses the methods registered in the dispatcher.

### Benefits

- The dispatcher could be easily repurposed to server also the `suave-as-a-contract` model.
- It solves the full lifecycle of writing the precompile (encoding, routing, gas requirement). With the current model, though encoding/decoding is solved, we need extra wiring to fix the rest.

## 📚 References

---


ferranbt:
Closed in favour of #94

PR #91, opened 2023-11-01
Optional  `BlockNumber` field in a bundle 

halo3mic:
## 📝 Summary

Propagate bundles with missing block-number field without block-number, instead of a block-number of value zero.

* Add `omitempty` to block-number field in `SBundle` and  `RpcSBundle`
* Make `BlockNumber` field of  `RpcSBundle` optional (pointer + Marshall/Unmarshall)

---


Ruteri (approved):

Ruteri:
Thanks!

PR #101, opened 2023-11-09
Bundle marshaling

Ruteri:
## 📝 Summary

* Adds TransactionArgs and Bundle types
* TransactionArgs are marshaled to rlp-encoded bytes, ready for signing and sending
* Bundle is marshaled to json that external builders expect
* Adds precompiles for encoding and decoding each of them

* [ ] Does not support signatures yet
* [ ] Big ints are encoded as uint64 which may overflow

## 📚 References

---


dmarzzz:
closing this in favor of doing it in suave-std https://github.com/flashbots/suave-std/issues/23

PR #103, opened 2023-11-16
Add new precompile to simulate a transaction and return complete output

ferranbt:
## 📝 Summary

This PR adds a new precompile `simulateTransaction` that does the same thing as `buildBlock` but instead of returning the profit of the block it returns a struct with multiple contextual information about the execution that includes the profit and the logs.

## 📚 References

---


ferranbt:
Done as part of #115

PR #110, opened 2023-11-24
Add precompile to make remote http calls

ferranbt:
## 📝 Summary

This PR adds a new precompile to make arbitrary GET/POST remote calls. The new precompile is `doHttpRequest` and takes a struct `HttpRequest` with the type:

```
struct HttpRequest {
        string url;
        string method;
        string[] headers;
        bytes body;
}
```

The way to call the precompile would be:
```
Suave.HttpRequest memory request;
request.url = "";
request.method = "GET"

bytes memory response = Suave.doHttpRequest(request);
```

## 📚 References

---


core/vm/contracts_suave.go
```diff
@@ -133,3 +136,42 @@ type suaveRuntime struct {
...
+	if request.Url == "" {
+		return nil, fmt.Errorf("url is empty")
+	}
+
+	var body io.Reader
+	if request.Body != nil {
+		body = bytes.NewReader(request.Body)
+	}
+
+	req, err := http.NewRequest(request.Method, request.Url, body)
```
Ruteri, line 153: We probably want to enforce a timeout right?

core/vm/contracts_suave.go
```diff
@@ -133,3 +136,42 @@ type suaveRuntime struct {
 }
 
 var _ SuaveRuntime = &suaveRuntime{}
+
+func (s *suaveRuntime) doHTTPRequest(request types.HttpRequest) ([]byte, error) {
+	if request.Method != "GET" && request.Method != "POST" {
+		return nil, fmt.Errorf("only GET and POST methods are supported")
+	}
+	if request.Url == "" {
```
Ruteri, line 144: I'd be a bit worried about allowing connecting to arbitrary urls

amiller: do you have some example targets for the requests in mind? anything that needs authentication, yiou'll have to at least set the request headers. I think this "http" library also allows https urls and applies a default certificate authority policy right?

core/vm/contracts_suave.go
ferranbt: Who decides the timeout? Another field in the request?

core/vm/contracts_suave.go
ferranbt: > I'd be a bit worried about allowing connecting to arbitrary URLs

Note that there is already another precompile which allows calling arbitrary URLs [here](https://github.com/flashbots/suave-geth/blob/main/suave/gen/suave_spec.yaml#L156).

We have not discussed yet the security model of making external requests. I was thinking of that to be a follow-up conversation since it is going to take several iterations to dial down. My initial take would be to have a flag in the Kettle to whitelist a list of domains. How does that sound?

> do you have some example targets for the requests in mind?

[This](https://github.com/flashbots/suave-geth/blob/feature/suave-std/suave/sol/suave-std/suavex.sol#L33) is an example. It makes a JSON-RPC Post request to simulate a transaction in the domain-specific node.

> I think this "http" library also allows https urls and applies a default certificate authority policy right?

Yes, it uses the Go standard "HTTP" library so it uses default certificate authorities.

core/vm/contracts_suave.go
metachris: should we set a sane default? 5s? 2s? 10s?

core/vm/contracts_suave.go
Ruteri: I guess you are correct

core/vm/contracts_suave.go
ferranbt: Added 5-second timeout as default for now.

Ruteri (approved):

PR #115, opened 2023-12-04
Stateless builder

ferranbt:
## 📚 References

---


core/vm/contracts_suave.go
```diff
@@ -196,3 +197,20 @@ func (s *suaveRuntime) doHTTPRequest(request types.HttpRequest) ([]byte, error)
 	}
 	return data, nil
 }
+
+func (s *suaveRuntime) newBuilder() (string, error) {
```
dmarzzz, line 201: would be interesting to think about two kettles sharing the same session ID and doing a `newBuilder(existingSessionID string)` or `connectBuilder`. We don't need to consider for this work just mentioning if it inspires anything again!

suave/builder/api/api_server.go
```diff
@@ -0,0 +1,43 @@
...
+
+func NewServer(s sessionManager) *Server {
+	api := &Server{
+		sessionMngr: s,
+	}
+	return api
+}
+
+type Server struct {
+	sessionMngr sessionManager
```
dmarzzz, line 23: should we have a mutex on this? seems possible you could call start a session from another SUAPP before the last new session call has been completed?

suave/builder/api/api_server.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func (s *Server) NewSession(ctx context.Context) (string, error) {
```
dmarzzz, line 26: wonder if we should consider some sort of sane session cap, additionally, should we add metrics to this maybe so we can observe?

suave/builder/builder.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+type builderConfig struct {
+	preState *state.StateDB
+	header   *types.Header
+	config   *params.ChainConfig
+	context  core.ChainContext
+}
+
+func newBuilder(config *builderConfig) *builder {
```
dmarzzz, line 28: @Ruteri curious your thoughts on if this is the best way to set up builder, I remember something non-obvious about reverts changing in builder recently so double check would be appreciated

suave/builder/session_manager.go
```diff
@@ -0,0 +1,138 @@
...
+	parent := s.blockchain.CurrentHeader()
+
+	chainConfig := s.blockchain.Config()
+
+	header := &types.Header{
+		ParentHash: parent.Hash(),
+		Number:     new(big.Int).Add(parent.Number, common.Big1),
+		GasLimit:   core.CalcGasLimit(parent.GasLimit, s.config.GasCeil),
+		Time:       1000,             // TODO: fix this
+		Coinbase:   common.Address{}, // TODO: fix this
```
dmarzzz, line 77: hmm yeah interesting to think what you would set this to. Two ideas come to mind:
- set to '0x + keccak256("SUAVE")` so its consistent for users 
- allow for the ability to set it. I think there may be some scenarios where you if your a builder who also has a contract controlled by your address that does something different based on profit in a block, then you would want to be able to set this. Happy to spell this out more but I think it's desirable.

core/vm/contracts_suave.go
dmarzzz: permisisons around this has some caution signs for me, you could mess with someones simulation if you knew their session ID if done naively

suave/gen/suave_spec.yaml
```diff
@@ -218,3 +236,20 @@ functions:
       fields:
         - name: response
           type: bytes
+  - name: newBuilder
+    address: "0x0000000000000000000000000000000053200001"
+    output:
+      fields:
+        - name: id
+          type: string
```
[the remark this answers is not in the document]
ferranbt: The session id is a `uuid` identifier. It gets rendered as a string.

suave/builder/session_manager.go
ferranbt: Agree, I think some of these params should be configurable when you start a session like the `extra` field.

suave/builder/api/api_server.go
ferranbt: The locks are on `sessionManager`.

suave/sol/libraries/Suave.sol
```diff
@@ -34,6 +34,19 @@ library Suave {
         bytes body;
     }
 
+    struct SimulateTransactionResult {
+        uint64 egp;
```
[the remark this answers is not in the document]
ferranbt: I would see that included as some type of getBalance call to the builder.

suave/builder/api/api_server.go
ferranbt: We could address this in a followup PR.

dmarzzz (approved):
LSTM

suave/builder/api/api_server.go
lthibault: #165

PR #116, opened 2023-12-04
use pending txs to build blocks

halo3mic:
### Motivation

Spend less Goerli ETH building Goerli blocks via Suave. 

### Solution 

Optionally include pending transactions while building a block via `suavex_buildEthBlock` and `suavex_buildEthBlockFromBundles`.

---------


dmarzzz:
> Sure! Just to clarify, usePending -> fillPending or commitPendingTxs -> fillPending?
correct! my thinking is roughly: usePending only means use pending transactions when actually what this is doing is filling the rest of the block with pending transactions

btw, do you mind adding this change to the [specs](https://github.com/flashbots/suave-specs/blob/main/specs/rigil/precompiles.md) as well, you might need to update precompile and the suave ex namespace section.

halo3mic:
> fillPending
> > Sure! Just to clarify, `usePending` -> `fillPending` or `commitPendingTxs` -> `fillPending`?
> 
> the first, my thinking is roughly: `usePending` only means use pending transactions when actually what this is doing is filling the rest of the block with pending transactions
> 
> btw, do you mind adding this change to the [specs](https://github.com/flashbots/suave-

Made the renaming changes. Regarding the spec update, I didn't find anything mentioning the composition of `BuildBlockArgs` ... not really confident about where exactly to add the struct definition in precompiles and suavex call.

ferranbt (approved):

dmarzzz:
@halo3mic can you rebase this?

halo3mic:
> @halo3mic can you rebase this?

Rebased!

PR #117, opened 2023-12-05
Introduce typed precompiles

ferranbt:
## 📝 Summary

This PR introduces counterparts to some of the current precompiles which take `[]byte` with typed objects.

- sendBundle: Sends a Bundle object following [this](https://docs.flashbots.net/flashbots-auction/advanced/rpc-endpoint#eth_sendbundle) spec.
- simulateTransactions: Simulates a Transaction object.
- sendMevShareBundle: Sends a mev share bundle object.
- encodeTxnRLP: Encodes a typed Transaction to RLP.

## 📚 References

---


dmarzzz:
@ferranbt are we still keen on this?

ferranbt:
> @ferranbt are we still keen on this?

I am keen on this but in combination with suave-std so I think I will do that first and then combine it with the changes in this PR.

PR #120, opened 2023-12-05
Add utilities to parse log events

ferranbt:
## 📝 Summary

This PR introduces a utility to parse event logs. An event in Ethereum has the form:
```
struct Event(
   uint64 num,
   uint64 indexed num2,
   uint256 num3
)
```
When serialized into JSON, indexed fields (e.g. num2) are included in an array of 32 bytes each called `topics`. Un-indexed fields are ABI encoded in a single `[]byte` field called `Data`.

Currently, there are only helper methods to separately parse the indexed/unindexed fields. This PR creates a new method `ParseLog` that combines both methods.

## 📚 References

---


dmarzzz (approved):
love it

PR #123, opened 2023-12-08
secp256k1 precompiles

zeroXbrock:
## 📝 Summary

- add randomUint precompile
- add secp256k1 precompiles
  - secp256k1Sign
  - secp256k1RecoverPubkey
  - secp256k1VerifySignature

~~TODO:~~

- [x] add tests

## 📚 References

---


core/vm/contracts_suave_runtime_adapter.go
```diff
@@ -42,14 +47,18 @@ var (
 	fetchBidsAddr                = common.HexToAddress("0x0000000000000000000000000000000042030001")
 	fillMevShareBundleAddr       = common.HexToAddress("0x0000000000000000000000000000000043200001")
 	newBidAddr                   = common.HexToAddress("0x0000000000000000000000000000000042030000")
+	randomUintAddr               = common.HexToAddress("0x00000000000000000000000000000000F00bA777")
```
metachris, line 50: should the hex addresses follow the same scheme as the others?

core/vm/contracts_suave.go
```diff
@@ -109,6 +112,28 @@ func (b *suaveRuntime) fetchBids(targetBlock uint64, namespace string) ([]types.
 	return bids, nil
 }
 
+func (b *suaveRuntime) randomUint() (*big.Int, error) {
```
metachris, line 115: should the max return value here be u256 or u64?

zeroXbrock: It should be u256 -- does golang's `big.Int` not go that high?

core/vm/contracts_suave_runtime_adapter.go
zeroXbrock: I see a loose pattern in how the addresses are chosen, but it doesn't seem totally consistent.
For example, from 
```go
fetchBidsAddr                = common.HexToAddress("0x0000000000000000000000000000000042030001")
```
we might infer that `0x00..42` implies MEV-Share bid-related precompiles. But then we see
```go
fillMevShareBundleAddr       = common.HexToAddress("0x0000000000000000000000000000000043200001")
```
So then we might infer that `0x00..42` is strictly bids and `0x00..43` is just bundle-related. But then we see
```go
simulateBundleAddr           = common.HexToAddress("0x0000000000000000000000000000000042100000")
```
So then my guess is that the scheme is "start with 0x000000000000000000000000000000004 and have some zeros in the middle?"... Is there an actual pattern I should be following?
I thought this address was fun :p but happy to change it -- just wondering how best to avoid conflicts w/ other precompile PRs.

core/vm/contracts_suave.go
metachris: `big.Int` is arbitrarily precision.

dmarzzz:
@zeroXbrock didnt realize my PR does the same #127 , Im curious why we need recover as a precompile if solidity already exposes ecrecover? also we dont include the curve name on signEthTransaction so Im not sure we need to for signing an arbitrary message either if its in ethereums native curve, only for non-native curves

dmarzzz:
closing because stale

PR #127, opened 2023-12-12
add signMessage precompile

dmarzzz:
## 📝 Summary

adds precompile to sign arbitrary digests using ECDSA signing key.

_Note: I added this precompile to `contracts_suave.go` since it is not ethereum specific._

## 📚 References

[SUAVE-spec PR](https://github.com/flashbots/suave-specs/pull/71)

---


ferranbt (dismissed):

ferranbt (commented):
Is there a conflict with #123?

dmarzzz:
> Is there a conflict with #123?

No different curve, and ECDSA recovery method is native to solidity so don't need all of them for this PR as in #123

dmarzzz:
yeah it's the same, I left a comment over there. I dont think we need to specify the curve in the precompile name

core/vm/contracts_suave.go
```diff
@@ -115,6 +116,20 @@ func (b *suaveRuntime) fetchDataRecords(targetBlock uint64, namespace string) ([
 	return records, nil
 }
 
+func (s *suaveRuntime) signMessage(digest []byte, signingKey string) ([]byte, error) {
```
ferranbt, line 119: Why is `signingKey` not bytes as well?

dmarzzz: the `signingKey` in our other precompiles are also strings. Im not entirely sure why we set this precedent but I believe it was @Ruteri who did

dmarzzz: https://github.com/flashbots/suave-specs/blob/main/specs/rigil/precompiles.md#signethtransaction

ferranbt (approved):

PR #109, opened 2023-11-23
Leak emitted logs during Confidential Request

ferranbt:
## 📝 Summary

Builds on top of #108. This PR "leaks" in the `SuaveTransaction` structure the logs emitted during the confidential request.

## 📚 References

---

* [x] I have seen and agree to CONTRIBUTING.md

PR #131, opened 2023-12-16
POC : Request results pattern

dmarzzz:
## 📝 Summary

This is a simple PoC of a logging pattern we could use to help display more information in the block explorer to help guide users on their SUAPPs internals.

We could then alter the block explorer to something like this (don't read into the screenshot too much, I simply googled "networking UI" and photo shopped it into a screen shot of the explorer)

Screenshot 2023-12-16 at 2 15 25 PM (2)

## 📚 References

---

### DIFF
--- a/core/vm/contracts_suave_eth.go
+++ b/core/vm/contracts_suave_eth.go
@@ -355,6 +355,13 @@ func executableDataToCapellaExecutionPayload(data *engine.ExecutableData) (*spec
 	}, nil
 }
 
+type SubmissionResults struct {
+	StatusCode    uint64
+	Destination   string
+	RoundTripTime *big.Int
+	Response      string
+}
+
 func (c *suaveRuntime) submitBundleJsonRPC(url string, method string, params []byte) ([]byte, error) {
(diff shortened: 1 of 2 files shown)


core/vm/contracts_suave_eth.go
```diff
@@ -355,6 +355,13 @@ func executableDataToCapellaExecutionPayload(data *engine.ExecutableData) (*spec
 	}, nil
 }
 
+type SubmissionResults struct {
+	StatusCode    uint64
+	Destination   string
+	RoundTripTime *big.Int
```
metachris, line 361: instead of bigInt complexity, perhaps go for uint64 in millis? or alternatively record `startTimeUnix` and `endTimeUnix` 🤔

dmarzzz: totally great suggestions! just threw together something quick to illustrate

dmarzzz:
using #109 we could actually just automatically leak this event on precompile calls. This of course forces the dapp into it. But the info leak is really more of a concern for the kettle as you can triangulate it's location by this (maybe not a bad thing!)

PR #144, opened 2024-01-06
Deprecate SuaveForge.sol

ferranbt:
## 📝 Summary

Now we have [better forge support](https://github.com/flashbots/suave-std/pull/11) on suave-std, we can deprecate SuaveForge.sol.

Let's merge this once we have documented the feature on `suave-std`.

## 📚 References

---


dmarzzz (approved):
LSTM

PR #145, opened 2024-01-06
Add forge status command

ferranbt:
## 📝 Summary

This PR adds a new `suave forge status` command which validates if there is a remote Suave node running or not. If the node is not running, it returns an error message with the prefix `not-ok:`. If the node is running, it returns empty. 

This feature is intended to be used with the `forge` integration to validate whether there is a node running. The error code is always 0 even for error messages. Otherwise, it complicates the integration because of the error handling of some of the cheatcodes.

## 📚 References

---

### DIFF
--- a/cmd/geth/forgecmd.go
+++ b/cmd/geth/forgecmd.go
@@ -15,12 +15,17 @@ import (
 	"github.com/urfave/cli/v2"
 )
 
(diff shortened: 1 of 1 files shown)


dmarzzz (approved):
LSTM

PR #146, opened 2024-01-08
Add goreleaser

ferranbt:
## 📝 Summary

This PR introduces `goreleaser` to build releases for `linux`, `darwin` and `windows` with `cgo` enabled. It also produces `docker` images and pushes them to `ghcr`.

## 📚 References

---

### DIFF
--- a/.github/workflows/releaser.yml
+++ b/.github/workflows/releaser.yml
@@ -1,0 +1,36 @@
+# .github/workflows/release.yml
+name: goreleaser
+
+on:
+  workflow_dispatch:
+  push:
+    tags:
+      - "*"
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: setup dependencies
+        uses: actions/setup-go@v2
+
+      - name: Login to GHCR
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Publich ref name
+        run: echo "Build for tag ${{ github.ref_name }}"
+
+      - name: release dry run
+        run: make release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TAG: ${{ github.ref_name }}
--- a/Makefile
+++ b/Makefile
@@ -54,3 +54,21 @@ devnet-down:
 
 fmt-contracts:
 	cd suave && forge fmt
+
+release:
+	# Generate the changelog for the release
+	./suave/scripts/parse-changelog.sh $(TAG) > /tmp/changelog.md
+
+	# Build and publish the release
+	docker run \
+		--rm \
+		-e CGO_ENABLED=1 \
+		-e GITHUB_TOKEN="$(GITHUB_TOKEN)" \
+		-v /tmp/changelog.md:/build/changelog.md \
+		-v /var/run/docker.sock:/var/run/docker.sock \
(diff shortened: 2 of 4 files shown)


.github/workflows/releaser.yml
```diff
@@ -0,0 +1,36 @@
...
+        uses: actions/setup-go@v2
+
+      - name: Login to GHCR
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Publich ref name
```
metachris, line 29: typo

.github/workflows/releaser.yml
```diff
@@ -0,0 +1,36 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: setup dependencies
+        uses: actions/setup-go@v2
+
+      - name: Login to GHCR
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
```
metachris, line 25: let's use docker hub, i'll help set this up

suave/scripts/parse-changelog.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
```
metachris, line 1: perhaps we should use https://github.com/orhun/git-cliff

metachris (approved):

metachris:
nice work!

Makefile
```diff
@@ -54,3 +54,21 @@ devnet-down:
...
+		-e CGO_ENABLED=1 \
+		-e GITHUB_TOKEN="$(GITHUB_TOKEN)" \
+		-v /tmp/changelog.md:/build/changelog.md \
+		-v /var/run/docker.sock:/var/run/docker.sock \
+		-v $(HOME)/.docker/config.json:/root/.docker/config.json \
+		-v `pwd`:/go/src/$(PACKAGE_NAME) \
+		-v `pwd`/sysroot:/sysroot \
+		-w /go/src/$(PACKAGE_NAME) \
+		ghcr.io/goreleaser/goreleaser-cross:v1.19.5 \
+		release --clean --release-notes /build/changelog.md
```
metachris, line 74: i think it'd be better to create a draft release here, and manually edit the description and publish when ready

.github/workflows/releaser.yml
metachris: wdyt?

probably docker hub is more common and easier to use, as you don't need to be logged into github (it's more easily accessible)

I've created https://hub.docker.com/r/flashbots/suave-geth and added env vars to the repo settings

.github/workflows/releaser.yml
metachris, line 29:
```suggestion
      - name: Public ref name
```

.github/workflows/releaser.yml
ferranbt: Did you set `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN`?

.github/workflows/releaser.yml
```diff
@@ -0,0 +1,36 @@
...
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Publich ref name
+        run: echo "Build for tag ${{ github.ref_name }}"
```
metachris, line 29: does this work? in the past I've had to use this way:

https://github.com/flashbots/mev-boost/blob/develop/.github/workflows/release.yaml#L20-L26

ferranbt: Yes, check this execution log https://github.com/ferranbt/suave-geth/actions/runs/7446372136/job/20256482872

.github/workflows/releaser.yml
ferranbt: Updated to `Log tag name`

PR #151, opened 2024-01-09
more suaveup readme

metachris:


ferranbt (approved):
Could you also update the makefile [here](https://github.com/flashbots/suave-geth/blob/main/Makefile#L18)?

metachris:
nice, done

PR #154, opened 2024-01-10
Add offchain package

lthibault:
## 📝 Summary
Begins formal separation of infrastructure and application logic.  
Introduces `offchain` package that defines a top-level `Env` struct.  This struct contains various APIs to off-chain services.  Currently this includes only `Datastore`.  Uses IPFS sidecar as test implementation.

Follow-up work:
1. Define precompiles to interact with `Datastore` from Solidity (in progress)
2. Migrate `cstore.Engine` to `Env`

## 📚 References

https://docs.ipfs.tech/concepts/content-addressing
https://github.com/multiformats/cid

---


metachris:
Noticed that the tests fail: https://github.com/flashbots/suave-geth/actions/runs/7471328612/job/20331422338?pr=154#step:9:2848

lthibault:
I wanted to add some additional context on the IPFS sidecar, as I don't think I've done a very good job of explaining myself so far (cc @Ruteri) 🙂

My reasons for proposing this change boil down to the following:
1. Running an IPFS sidecar is very helpful for me to prototype fast.
2. I am adding a feature flag to enable this, so it won’t interfere with any dev or prod setups.
3. We don’t have to keep IPFS.  We can discuss when the time comes to enable the feature in prod.

If the above sounds reasonable to you guys (cc @ferranbt ), here is the course I have plotted for myself:

- [ ] Fix the panic in the unit tests.  (WIP)
- [ ] Await one final review from you guys (possibly including a quick sync-up call?)
- [ ] Proceed with follow-up PRs

lthibault:
Okay, I think this is ready to go, pending approval.

I'm proceeding with the next PR:  implementing precompiles to interact with the datastore.

suave/offchain/blockstore.go
```diff
@@ -0,0 +1,32 @@
...
+	"io"
+
+	"github.com/ipfs/boxo/path"
+	"github.com/ipfs/go-cid"
+	iface "github.com/ipfs/kubo/core/coreiface"
+	"github.com/ipfs/kubo/core/coreiface/options"
+	"github.com/multiformats/go-multihash"
+)
+
+type Blockstore struct {
```
dmarzzz, line 14: why is this called blockstore? particularly asking because in a blockchain client it makes it sound like its for storing blocks

metachris: agree, good point

suave/offchain/offchain.go
```diff
@@ -0,0 +1,26 @@
+package offchain
+
+import (
+	"github.com/ipfs/kubo/client/rpc"
+	iface "github.com/ipfs/kubo/core/coreiface"
+)
+
+type Env struct {
```
metachris, line 8: `Env` feels a little non-descriptive, as in it doesn't really convey a meaning (offchain.Env) 🤔 

What is this for, eventually?

suave/offchain/blockstore.go
lthibault: Yeah, it's a bad choice 😄  I'll rename it to something like `Blobstore`.

suave/offchain/offchain.go
lthibault: It's meant to contain the "offchain environment", i.e. the collection of APIs and services that we export to SUAPPs for off-chain data/computation.  Ordinarily, I'd agree that `Env` doesn't convey a lot of meaning, but I think it actually is an "environment" in this case.

PR #158, opened 2024-01-11
Add blockstore precompiles

lthibault:
⚠️ Stacks on top of #154.  Merge that first.

## 📝 Summary

Adds two precompiles for interacting directly with the datastore (tentatively renamed to `Blockstore`, though this may change).

## 📚 References

---


lthibault:
Merge conflict is noted.  Will resolve once #154 is merged in.

metachris:
should this PR point to the branch of #154 instead of `main`? not sure

core/vm/offchain.go
```diff
@@ -0,0 +1,9 @@
+package vm
+
+import (
+	"github.com/ethereum/go-ethereum/suave/offchain"
+)
+
+type System interface {
+	Blocks() offchain.Blockstore
```
metachris, line 8: why is this using `Blocks` and `Blockstore`? wouldn't this be kind of a generic store? the name implies it's for blocks 🤔

eth/backend.go
```diff
@@ -319,6 +321,11 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 	stack.RegisterLifecycle(eth)
 	stack.RegisterLifecycle(confidentialStoreEngine)
 
+	// TODO(lthibault):  replace this with a proper CLI flag.
+	if os.Getenv("SUAVE_EXPERIMENTAL_IPFS") != "" {
+		stack.RegisterLifecycle(&offchain.Env{})
```
metachris, line 326: ideally add a log statement here so it's clearly visible on startup that this is indeed enabled

core/vm/offchain.go
lthibault: Yeah, it's confusing.  I'll change this to something like `Blob`, which is more accurate.

lthibault:
> should this PR point to the branch of #154 instead of `main`? not sure

@metachris I was hoping to merge these sequentially into `main`, to build incrementally.

PR #164, opened 2024-01-17
EIP-712 messages as confidential requests

ferranbt:
## 📚 References

---


internal/ethapi/api.go
```diff
@@ -1908,6 +1908,50 @@ func (s *TransactionAPI) FillTransaction(ctx context.Context, args TransactionAr
 	return &SignTransactionResult{data, tx}, nil
 }
 
+func (s *TransactionAPI) SendRawTransaction2(ctx context.Context, eip712Envelope *types.ConfidentialComputeRequest2, input hexutil.Bytes) (common.Hash, error) {
+	// Entrypoint for signed eip-712 messages
```
ferranbt, line 1912: TODO: Verify that the message and the signature match.

suave/sdk/sdk.go
```diff
@@ -64,29 +68,76 @@ func (c *Contract) SendTransaction(method string, args []interface{}, confidenti
...
+	typedData := builder.Build(record)
+
+	typedDataHashed, err := typedData.Hash()
 	if err != nil {
-		return nil, err
+		panic(err)
 	}
 
-	computeRequestBytes, err := computeRequest.MarshalBinary()
+	signedMsg, err := c.client.Sign(typedDataHashed[:])
```
ferranbt, line 113: So this does not work. `suaveTransaction` uses the signature values from record (`v`, `r`, `s` and `chainid`) to validate the transaction. But, those fields are not populated.
But, can they be populated? Not really, because you are signing the EIP-712 message and not the record. So, I think we have to move the signature values to the outer envelope of the EIP-712, thoughts?

core/types/confidential.go
```diff
@@ -89,6 +91,97 @@ func (tx *ConfidentialComputeRecord) setSignatureValues(chainID, v, r, s *big.In
 	tx.ChainID, tx.V, tx.R, tx.S = chainID, v, r, s
 }
 
+type ConfidentialComputeRequest2 struct {
+	EIP712    *eip712.EIP712TypedData `json:"eip712"`
+	Signature []byte                  `json:"signature"`
```
ferranbt, line 96: Signature values in the record are deprecated because you are signing the EIP-712 message, not the record.

core/types/confidential.go
```diff
@@ -89,6 +91,97 @@ func (tx *ConfidentialComputeRecord) setSignatureValues(chainID, v, r, s *big.In
...
+	Sender    common.Address          `json:"sender"`
```
ferranbt, line 97: This is hacky thing because Geth expects all the signatures in the v, r, s form but in this case is just []byte.

core/types/confidential.go
```diff
@@ -89,6 +91,97 @@ func (tx *ConfidentialComputeRecord) setSignatureValues(chainID, v, r, s *big.In
 	tx.ChainID, tx.V, tx.R, tx.S = chainID, v, r, s
 }
 
+type ConfidentialComputeRequest2 struct {
```
ferranbt, line 94: This is a transaction type, we could say that is the EIP-712 transaction type. Record is only the message type, like the Suave intent-format.

core/types/transaction_signing.go
```diff
@@ -289,13 +290,18 @@ func (s suaveSigner) Sender(tx *Transaction) (common.Address, error) {
...
 		ccr = &txdata.ConfidentialComputeRecord
 
 		if txdata.ConfidentialInputsHash != crypto.Keccak256Hash(txdata.ConfidentialInputs) {
 			return common.Address{}, errors.New("confidential inputs hash mismatch")
 		}
 	case *ConfidentialComputeRecord:
+		fmt.Println("CCXX")
 		ccr = txdata
+	case *ConfidentialComputeRequest2:
+		// BIG TODO: signature validation, forget about v, r and s, just use the signautre field?
```
ferranbt, line 303: Skip signature check

suave/sdk/sdk.go
```diff
@@ -64,29 +68,80 @@ func (c *Contract) SendTransaction(method string, args []interface{}, confidenti
...
+		KettleAddress:          c.client.kettleAddress,
+		Nonce:                  nonce,
+		To:                     &c.addr,
+		Value:                  big.NewInt(0),
+		GasPrice:               gasPrice,
+		Gas:                    1000000,
+		Data:                   calldata,
+		ConfidentialInputsHash: crypto.Keccak256Hash(confidentialDataBytes), // ?
+
+		// initialize empty fields
```
ferranbt, line 81: Initialise them because the EIP-712 library cannot handle nils.

suave/sdk/sdk.go
```diff
@@ -64,29 +68,80 @@ func (c *Contract) SendTransaction(method string, args []interface{}, confidenti
...
+		computeRequest, err := types.SignTx(types.NewTx(&types.ConfidentialComputeRequest{
+			ConfidentialComputeRecord: *record,
+			ConfidentialInputs:        confidentialDataBytes,
+		}), signer, c.client.key)
+		if err != nil {
+			return nil, err
+		}
+	*/
+
+	domain := &eip712.EIP712Domain{
```
ferranbt, line 98: Feedback on this fields is welcomed.

suave/sdk/sdk.go
```diff
@@ -64,29 +68,80 @@ func (c *Contract) SendTransaction(method string, args []interface{}, confidenti
...
+		Name:              "suave?",
+		Version:           "1.0",
+		VerifyingContract: "0x0000000000000000000000000000000000000000", // ??
```
zeroXbrock, line 101: this is included in the eip712 spec because these signatures are typically validated by a smart contract; the contract would check that `msg.verifyingContract == address(this)`.
Since we're validating signatures offchain, I think this can be ignored and filled with a dummy value

suave/sdk/sdk.go
```diff
@@ -64,29 +68,80 @@ func (c *Contract) SendTransaction(method string, args []interface{}, confidenti
...
+		ChainId:           big.NewInt(1),                                // suave chain id?
```
zeroXbrock, line 102: yes, since the primary message is for suave

suave/sdk/sdk.go
```diff
@@ -64,29 +68,80 @@ func (c *Contract) SendTransaction(method string, args []interface{}, confidenti
...
+		Nonce:                  nonce,
+		To:                     &c.addr,
+		Value:                  big.NewInt(0),
+		GasPrice:               gasPrice,
+		Gas:                    1000000,
+		Data:                   calldata,
+		ConfidentialInputsHash: crypto.Keccak256Hash(confidentialDataBytes), // ?
+
+		// initialize empty fields
+		ChainID: big.NewInt(1337), // suave chain id?
```
zeroXbrock, line 82: yep, same as https://github.com/flashbots/suave-geth/pull/164/files#r1459746595

suave/sdk/sdk.go
```diff
@@ -64,29 +68,80 @@ func (c *Contract) SendTransaction(method string, args []interface{}, confidenti
...
+		V:       big.NewInt(0),
```
zeroXbrock, line 83: do/should we have a version of this struct without v,r,s?

zeroXbrock:
good first stab! so the client builds a CCRecord and encodes it into an EIP-712 payload , then signs and sends that to the `eth_sendRawTransaction2` endpoint in a CCRequest2 envelope with confidentialInputs appended. I think this'll work. Just needs signature verification.

PR #165, opened 2024-01-18
Limit number of open sessions

### LINKED ISSUE #153: Max Builder Sessions
# Rationale

https://github.com/flashbots/suave-geth/pull/115#discussion_r1431376986

lthibault:
## 📝 Summary

Use a semaphore channel to limit the number of open sessions. This is a hack because we're relying on session timeout behavior. Recommend ref-counting API that releases refs when block-height changes in near-term future.

Resolves #153 

## 📚 References

---

### DIFF
--- a/suave/builder/api/api_server.go
+++ b/suave/builder/api/api_server.go
@@ -6,13 +6,13 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 )
 
-// sessionManager is the backend that manages the session state of the builder API.
-type sessionManager interface {
-	NewSession() (string, error)
+// SessionManager is the backend that manages the session state of the builder API.
+type SessionManager interface {
+	NewSession(context.Context) (string, error)
 	AddTransaction(sessionId string, tx *types.Transaction) (*types.SimulateTransactionResult, error)
 }
 
-func NewServer(s sessionManager) *Server {
+func NewServer(s SessionManager) *Server {
 	api := &Server{
 		sessionMngr: s,
 	}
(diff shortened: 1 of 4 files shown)


suave/builder/api/api_server.go
```diff
@@ -6,25 +6,25 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 )
 
-// sessionManager is the backend that manages the session state of the builder API.
-type sessionManager interface {
-	NewSession() (string, error)
+// SessionManager is the backend that manages the session state of the builder API.
```
lthibault, line 9: I noticed that this was an unexported type being returned from the public function `NewServer`, which is kinda awkward.  Exporting `SessionManager` is a drive-by improvement.

lthibault:
Test passes locally ... re-running 🧐 
I think this is ready for review in the meantime, though.

lthibault:
I'm noticing that failing test is [potentially non-deterministic](https://github.com/flashbots/suave-geth/blob/main/suave/cstore/redis_backends_test.go#L58-L63).  Did I break this, or is something flaky?

Test passes on my laptop, so I can't `git bisect` to see if I introduced this 🤔

metachris:
nice code and comments. is there a good way to write a test ensuring it works as expected?

lthibault:
Added unit tests.  @metachris I think it's a good idea to hook in a envvar flag here, but I'd like to do it in a separate PR because it'll affect a different set of modules.  Merging now and following up with that change.

PR #172, opened 2024-01-24
return valid value and error on suave runtime adapter

jinmel:
## 📚 References

---


core/vm/contracts_suave_runtime_adapter.go
```diff
@@ -134,7 +134,7 @@ func (b *SuaveRuntimeAdapter) buildEthBlock(input []byte) (res []byte, err error
 	unpacked, err = artifacts.SuaveAbi.Methods["buildEthBlock"].Inputs.Unpack(input)
 	if err != nil {
 		err = errFailedToUnpackInput
-		return
+		return nil, err
```
ferranbt, line 137: It uses named values in the function signature
```
func (b *SuaveRuntimeAdapter) buildEthBlock(input []byte) (res []byte, err error) {
```
and then the value is reused
```
unpacked, err =
```

jinmel: thanks for teaching me about this syntax!

PR #173, opened 2024-01-24
Generic buildEthblock precompile method

jinmel:
## 📝 Summary

As the buildEthblock precompile is opinionated with mev-share, this patch strips mev-share related namespaces from `buildEthBlock` to make it more generic 

## 📚 References

---


ferranbt (changes requested):
We have to hold on this change for now because some integration tests use this precompile.

dmarzzz:
yes `buildEthBlock` is unfortunately opinionated with MEV-share but we can't just remove it and then not offer the existing functionality anywhere 😅

PR #178, opened 2024-01-31
Use call instead of delegate call in precompile calls

ferranbt:
## 📝 Summary

This PR replaces the `staticCall` with a normal `call` when a Suave precompile is called. This does not affect the security model since precompiles cannot access state anyway (`staticCall` reverts if the contract tries to update a state variable).

But, it is helpful for the forge integration to make a mock ConfidentialStore in Solidity and use it as the target of the precompiles.

## 📚 References

---


lthibault (approved):
Send it 👍

PR #185, opened 2024-02-02
Add local mode execution

ferranbt:
## 📝 Summary

This PR adds a `--local` flag to the `suave-geth forge` command. If the flag is set, the command will resolve the precompile logic locally and it will not try to call a running `suave-geth` instance.

This works for all the precompiles except for the confidential store ones since they are stateless. If any of those precompiles is executed in this local mode, it will return an error.

The PR also includes flags to configure all the possible context of the execution of the precompiles (i.e. whitelist, eth backend address, etc...).

Also, the configuration can be set from a configuration file in toml following the `profile` convention from Foundry files.

```
[profile.suave]
whitelist = ["*"]
eth_backend = "http:..."
```

## 📚 References

---


lthibault:
Quick top-level question as I dive into this:  what is the intended use-case for this feature?  Local dev? Testing?

cmd/geth/forgecmd.go
```diff
@@ -1,28 +1,129 @@
...
+
+	// create the suave context
+	var suaveEthBackend suave.ConfidentialEthBackend
+	if cfg.EthBackend != "" {
+		suaveEthBackend = suave_backends.NewRemoteEthBackend(cfg.EthBackend)
+	} else {
+		suaveEthBackend = &suave_backends.EthMock{}
+	}
+
+	ecdsaKey, _ := crypto.GenerateKey()
```
lthibault, line 99: Can we check these errors?  It's fine to panic if we "know" they'll never fail.  I've just been bitten too many times by unchecked errors and have become a little neurotic as a result.

cmd/geth/forgecmd_test.go
```diff
@@ -0,0 +1,54 @@
...
+	for _, f := range flags {
+		if err := f.Apply(set); err != nil {
+			t.Fatal(err)
+		}
+	}
+	set.SetOutput(io.Discard)
+	return set
+}
+
+func TestForgeReadConfig(t *testing.T) {
```
lthibault, line 25: Since this is a new test, I'd recommend `t.Parallel()`.

lthibault (changes requested):

ferranbt:
> Quick top-level question as I dive into this: what is the intended use-case for this feature? Local dev? Testing?

This is the feature that `suave-std` uses to execute the precompiles with forge. When Solidity tests in forge are executed, the VM that is used (anvil) does not have the precompiles from Suave. Thus, we add some Solidity "Connectors" in those addresses that make external calls to this binary to resolve the logic. This only works for stateless executions.

lthibault (approved):
LGTM!  Thanks for clarifying the purpose -- makes sense.

PR #186, opened 2024-02-03
Feature/max concurrent builder session

allnil:
Draft adding env var configuration for SessionManager
Will play with it in this PR,
started with Viper and config init from the environment with Viper


allnil:
I made it work but I don't know how idiomatic it is in the context of the whole repo.
Will appreciate any comments and ideas on how to make it nit!

ferranbt:
Thank you for your contribution! Two comments from my side. First, we are moving the backend builder implementation to an independent repository in [suave-execution-geth](https://github.com/flashbots/suave-execution-geth). Could you make the PR there?

Also, we are always very mindful when adding new dependencies so I would prefer not to add a new one (`vyper`) just to handle the configuration of the builder, just `os.GetEnv` would be fine by me.

allnil:
@ferranbt 
got it! thank you!

Issue #176, opened 2024-01-25
No Deneb payload


halo3mic:
When submitting the block to the relay via Suave I get bad request with error `{"code":400,"message":"not deneb payload"}\n`.

The source of the error is [here in the relay code](https://github.com/flashbots/mev-boost-relay/blob/d2237f875f60363959db03b5114141fa16a685b2/services/api/service.go#L1648-L1653).

It appears suave-geth and suave-execution-geth need support for [the new payload](https://github.com/flashbots/builder/blob/677a32d6ea01d1500458d594ff6c7d41b2b315c9/builder/builder.go#L637-L671) with blobs.

lthibault:
@halo3mic Are you sending SSZ or JSON encoded payloads?  Can you provide an example payload so that we can reproduce?

lthibault:
@halo3mic Okay, I think I've narrowed it down a bit.  Any chance you'd be willing to take over and submit a PR? 😃 

I believe the issue is in `buildEthBlock`, [here](https://github.com/flashbots/suave-geth/blob/main/core/vm/contracts_suave_eth.go#L262-L266).  Notice that it's building a Capella block submission request.  Presumably that should be a Deneb variant.  I believe the `attestantio/go-builder/client` library may need a version bump.

Wanna give it a shot?

lthibault:
Per conversation on Telegram, #188 has not resolved the issue.

lthibault:
I suspect the issue might be [here](https://github.com/flashbots/suave-geth/blob/main/miner/worker.go#L1298). The `beaconRoot` is not included as part of the block (see equivalent here in the [normal builder](https://github.com/flashbots/builder/blob/deneb/miner/worker.go#L1118)).

There are actually are three new fields introduced by Deneb:

- `ParentBeaconRoot `
- `ExcessBlobGas`
- `BlobGasUsed`

lthibault:
Okay, I think the beacon root needs to be set, as `beaconRoot` is part of the [`Header` struct](https://github.com/flashbots/builder/blob/deneb/core/types/block.go#L95) when calculating the block hash.

PR #192, opened 2024-02-12
Add BeaconHash, ExcessBlobGas and BlobGasUsed to BuildBlockArgs.

lthibault:
## 📝 Summary

Attempt to fix https://github.com/flashbots/suave-geth/issues/176 by taking in `BeaconHash`, `ExcessBlobGas` and `BlobGasUsed` as arguments from the SUAPP.

Note that we are running behind the latest Deneb changes in upstream Geth, so it is possible that this fix is not sufficient.  In particular, I am uncertain as to whether the (de)serialization of arguments will work.  Input on this issue is most welcome.

As we don't have a great way of testing this right now, I would recommend merging and seeing if this solves the reported issue.  In the meantime, I am hoping that CI might helpfully fail.

Lastly, I've gone ahead and modified some files that were created with go-generate.  @ferranbt, do you know where/how I can do this the Right Way?

---


lthibault:
P.S.:  I'll need to make corresponding changes to `suave-execution-geth` before this can be tested.

lthibault:
Per Slack discussion, SUAVE chain doesn't require blob support, so reverting changes to block and consensus types.

lthibault:
I'm getting a very cryptic failure in CI.  I've tried stepping through with a debugger, at which point the bug doesn't reproduce.  Oh boy...

Thinking about this for a moment, my first thought is that this may have to do with payload unmarshalling on the remote end?  Any thoughts?

lthibault:
Note that this might fail with a hash mismatch in cases where the block args are nil, which I don’t think will happen, since a SUAPP should be passing in params… but I’m not 100% sure.

suave/sol/libraries/Suave.sol
```diff
@@ -218,8 +219,8 @@ library Suave {
         return abi.decode(data, (DataRecord));
     }
 
-    function privateKeyGen() internal view returns (string memory) {
-        (bool success, bytes memory data) = PRIVATE_KEY_GEN.staticcall(abi.encode());
+    function privateKeyGen() internal returns (string memory) {
+        (bool success, bytes memory data) = PRIVATE_KEY_GEN.call(abi.encode());
```
dmarzzz, line 223: i believe this is on purpose?

dmarzzz (approved):

suave/sol/libraries/Suave.sol
lthibault: I didn't change it; this is the output of Ferran's codegen tool.  I do believe there was a recent PR to make this a dynamic call, though.

lthibault: Resolving and merging, since you approved :)

PR #194, opened 2024-02-14
Introduce a new engine for the local confidential store that does not have security guarantees

ferranbt:
## 📝 Summary

This PR adds another flavour of the local confidential store that does not have any authentication check. This is because we are moving (at least in the Andromeda track) towards using the application layer to do the validation checks leveraging SGX.

## 📚 References

---


suave/cstore/engine2.go
```diff
@@ -0,0 +1,283 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/log"
+	suave "github.com/ethereum/go-ethereum/suave/core"
+	"github.com/google/uuid"
+
+	"github.com/gorilla/mux"
+)
+
+type Engine2 struct {
```
lthibault, line 22: Can we give this a more descriptive name?  Maybe `NonAuthenticatedEngine`?  I think a docstring might be helpful, too.

Ditto for `NewEngine2`.

suave/cstore/engine2.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+	log.Info("Server shut down gracefully")
+	return nil
+}
+
+func (a *Engine2) finalize(store *TransactionalStore2) error {
+	for _, record := range store.pendingRecords {
+		err := a.storage.InitRecord(record)
+		if err != nil {
+			// TODO: deinitialize!
```
lthibault, line 84: Is this TODO something we need to do before merging?

suave/cstore/transactional_store2.go
```diff
@@ -0,0 +1,125 @@
...
+	"errors"
+	"fmt"
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	suave "github.com/ethereum/go-ethereum/suave/core"
+)
+
+type TransactionalStore2 struct {
```
lthibault, line 13: Same as with `Engine2`.  I think a more descriptive name would be very helpful.

suave/cstore/engine2.go
```diff
@@ -0,0 +1,283 @@
...
+
+	return a
+}
+
+func (a *Engine2) Start() error {
+	log.Info("Server started on :8080")
+
+	go func() {
+		if err := a.server.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			log.Error("Error starting server: %s\n", err)
```
lthibault, line 63: I'm noting some inconsistency in capitalization in log messages.  I believe the convention in Go (and most languages) is not to capitalize log strings.  I would suggest doing the same here.

I'll flag the other locations in the PR.

suave/cstore/engine2.go
```diff
@@ -0,0 +1,283 @@
...
+	return nil
+}
+
+func (a *Engine2) Stop() error {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	if err := a.server.Shutdown(ctx); err != nil {
+		return err
+	}
+	log.Info("Server shut down gracefully")
```
lthibault, line 76: capitalization

suave/cstore/engine2.go
```diff
@@ -0,0 +1,283 @@
...
+func NewEngine2(storage ConfidentialStorageBackend) *Engine2 {
+	a := &Engine2{
+		storage:  storage,
+		sessions: make(map[string]*TransactionalStore2),
+	}
+
+	r := mux.NewRouter()
+
+	r.NotFoundHandler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		http.Error(w, "Not found", http.StatusNotFound)
```
lthibault, line 39: Lowercase (see note about logging below - applies here too, I think)

suave/cstore/engine2.go
```diff
@@ -0,0 +1,283 @@
...
+	key := mux.Vars(r)["key"]
+	dataId := decodeDataId(r)
+
+	session := a.getSession(id)
+	if session == nil {
+		http.Error(w, "session not found", http.StatusNotFound)
+		return
+	}
+
+	val, err := session.Retrieve(dataId, common.Address{}, key)
```
lthibault, line 184: Is the empty address intentional?

lthibault (changes requested):

suave/cstore/engine2.go
ferranbt: All the log prints are capitalised in Geth https://github.com/search?q=repo%3Aflashbots%2Fsuave-geth%20log.Print&type=code. Do you refer to error messages or prints on the console?

suave/cstore/engine2.go
ferranbt: This was an argument required for the storage implementation which in turn was not used. We can do a followup PR to change the storage interface. For this PR I wanted to have the minimum possible changes to the legacy store.

PR #195, opened 2024-02-14
add params BlobGasUsed and ExcessBlobGas to block bid

halo3mic:


halo3mic:
Without this the block hash check fails in builder's ExecutionPayloadV3ToBlock/ExecutableDataToBlock ([link](https://github.com/flashbots/builder/blob/677a32d6ea01d1500458d594ff6c7d41b2b315c9/beacon/engine/types.go#L390-L434)).

password
```diff
@@ -0,0 +1 @@
+admin
```
lthibault, line 1: Was this file meant to be added?

halo3mic: Deleted!

lthibault (approved):

PR #196, opened 2024-02-15
SUAVE dns

jinmel:
## 📝 Summary

Introduces SuaveEthRemoteBackendEnpointV2 flag to support multi domain kettle. `suave.eth.remote-endpoint` now accepts rpc endpoint configuration formatted in `chain_id=rpc_url`

## 📚 References
Generalizing suave execution namespace for multiple chains:  https://github.com/flashbots/suave-specs/issues/79

---


jinmel:
Should I keep the backwards compatibility by introducing new precompiles with different signatures?  (i.e. buildEthBlockV2, simulateBundleV2 ...) It seems like the integration tests are failing due to changes in precompiles

ferranbt (changes requested):
Overall looks good, could you write a unit test for the `contracts_suave.go` change and then another e2e integration? (`suave/e2e`).

cmd/geth/forgecmd.go
```diff
@@ -36,6 +36,10 @@ var (
 		Name:  "whitelist",
 		Usage: `The whitelist external endpoints to call`,
 	}
+	dnsRegistryForgeFlag = &cli.StringSliceFlag{
```
ferranbt, line 39: Love it!

jinmel:
@ferranbt Yeah I was also wondering if the kettle should set the port as well but that wouldn't be dns anymore. And also, if we are thinking of 
```solidity
Suavex.simulateTxn("goerlisuavex", ...)
```
we should call this endpoint aliasing rather than dns.

jinmel:
@ferranbt After giving it some thought, I believe it's a better practice for developers if we resolve an alias name directly to its full URL, port included. It simplifies things by hiding the complexities of the URL endpoint. Let me know if you think otherwise

core/vm/contracts_suave.go
```diff
@@ -220,12 +220,20 @@ func (s *suaveRuntime) doHTTPRequest(request types.HttpRequest) ([]byte, error)
...
-	for _, allowedDomain := range s.suaveContext.Backend.ExternalWhitelist {
-		if allowedDomain == "*" || allowedDomain == parsedURL.Hostname() {
-			allowed = true
-			break
+	// resolve dns if possible
+	if endpoint, ok := s.suaveContext.Backend.DnsRegistry[request.Url]; ok {
+		request.Url = endpoint
+		allowed = true
+	} else {
+		// check if the domain is allowed
```
ferranbt, line 228:
```suggestion
parsedURL, err := url.Parse(request.Url)
if err != nil {
     panic(err)
}
```
You do not have to parse the url if its a dns domain.

core/vm/contracts_suave.go
```diff
@@ -220,12 +220,20 @@ func (s *suaveRuntime) doHTTPRequest(request types.HttpRequest) ([]byte, error)
 	}
 
 	var allowed bool
-	for _, allowedDomain := range s.suaveContext.Backend.ExternalWhitelist {
-		if allowedDomain == "*" || allowedDomain == parsedURL.Hostname() {
-			allowed = true
-			break
+	// resolve dns if possible
```
ferranbt, line 223: Line 217, do you have to parse the url? I guess that is only required if not a dns entry.

ferranbt (approved):
Only one comment to resolve but it is minor since the tests work.

core/vm/contracts_suave.go
jinmel: good point!

PR #197, opened 2024-02-15
Fix dancun

halo3mic:
## About 
* This is POC for building blocks with blob transactions.
* Relates to PR https://github.com/flashbots/suave-execution-geth/pull/18
* For testing purposes it includes sending block to relay while block building - this should be removed

## Example of successful blocks with this version (and related external PR)
* [10543547](https://goerli.etherscan.io/block/10543547) see beacon explorer for 5 included blobs [here](https://goerli.beaconcha.in/block/10543547#blobs)
* [10543699](https://goerli.etherscan.io/block/10543547) see beacon explorer for 5 included blobs [here](https://goerli.beaconcha.in/block/10543699#blobs)


halo3mic:
What I noticed is that `BUILD_ETH_BLOCK` precompile always reverts with no error msg for some reason and the block was never included on-chain. I suspected that this could be gaslimit issue due to the size of blobs and put this snippet in  `buildEthBlock`. Now the precompile still reverts, like before, but the block is included on-chain like the example blocks in the previous comment. 

```go
res, err := b.submitEthBlockToRelay("https://0xafa4c6985aa049fb79dd37010438cfebeb0f2bd42b115b89dd678dab0670c1de38da0c4e9138c9290a398ecd9a0b3110@boost-relay-goerli.flashbots.net", bidBytes)
if err != nil {
    return nil, nil, fmt.Errorf("could not submit block to relay: %w", err)
}
log.Info("submitted block to relay", "response", string(res))
```

The snippet only serves for testing purposes and should be removed. Likewise, this snippet is not needed if gaslimit of local client is raised sufficiently, however didn't have any luck doing that.

PR #199, opened 2024-02-18
Extend the keygen and sign precompiles to handle multiple schemes

ferranbt:
## 📝 Summary

This PR overloads the `signMessage` and `generateKey` precompiles with an enum flag that determines the type of signature to use. Right now it provides two signature schemes, SECP256 and BLS.

## 📚 References

---


dmarzzz (approved):
LSTM - can we update https://github.com/flashbots/suave-specs/tree/main/specs/rigil also plz

PR #200, opened 2024-02-19
Add nightly releases

ferranbt:
## 📝 Summary

This PR enables nightly releases for `suave-geth`.

## 📚 References

---

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -97,10 +97,10 @@ dockers:
     image_templates:
       - "flashbots/suave-geth:{{ .ShortCommit }}-amd64"
       - "flashbots/suave-geth:v{{ .Version }}-amd64"
-      {{- if not .IsNightly }}
-      - flashbots/suave-geth:latest-amd64
-      {{ else }}
+      {{- if .IsNightly }}
       - flashbots/suave-geth:latest-nightly-amd64
+      {{ else }}
+      - flashbots/suave-geth:latest-amd64
       {{- end }}
   - dockerfile: ./Dockerfile.suave
     use: buildx
@@ -111,10 +111,10 @@ dockers:
     image_templates:
       - "flashbots/suave-geth:{{ .ShortCommit }}-arm64"
       - "flashbots/suave-geth:v{{ .Version }}-arm64"
-      {{- if not .IsNightly }}
-      - flashbots/suave-geth:latest-arm64
-      {{ else }}
+      {{- if .IsNightly }}
       - flashbots/suave-geth:latest-nightly-arm64
+      {{ else }}
+      - flashbots/suave-geth:latest-arm64
       {{- end }}
 
 docker_manifests:
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -97,7 +97,11 @@ dockers:
     image_templates:
       - "flashbots/suave-geth:{{ .ShortCommit }}-amd64"
       - "flashbots/suave-geth:v{{ .Version }}-amd64"
-      - "{{ if not .IsNightly }}flashbots/suave-geth:latest-amd64{{ end }}"
+      {{- if not .IsNightly }}
+      - flashbots/suave-geth:latest-amd64
+      {{ else }}
+      - flashbots/suave-geth:latest-nightly-amd64
+      {{- end }}
   - dockerfile: ./Dockerfile.suave
     use: buildx
     goarch: arm64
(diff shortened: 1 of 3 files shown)


.goreleaser.yaml
```diff
@@ -96,8 +96,8 @@ dockers:
       - --platform=linux/amd64
     image_templates:
       - "flashbots/suave-geth:{{ .ShortCommit }}-amd64"
-      - "flashbots/suave-geth:{{ .Tag }}-amd64"
-      - "flashbots/suave-geth:latest-amd64"
+      - "flashbots/suave-geth:v{{ .Version }}-amd64"
```
ferranbt, line 99: It was using Tag before, but it needs to use Version because that is the field that the `nightly` section overrides during nightly releases.

.goreleaser.yaml
```diff
@@ -123,6 +123,12 @@ docker_manifests:
       - "flashbots/suave-geth:latest-amd64"
```
ferranbt, line 123: Not sure If I have to do something with this.

.goreleaser.yaml
metachris, line 99: also tag it `:latest-nightly`?

.goreleaser.yaml
```diff
@@ -96,8 +96,8 @@ dockers:
       - --platform=linux/amd64
     image_templates:
       - "flashbots/suave-geth:{{ .ShortCommit }}-amd64"
-      - "flashbots/suave-geth:{{ .Tag }}-amd64"
-      - "flashbots/suave-geth:latest-amd64"
+      - "flashbots/suave-geth:v{{ .Version }}-amd64"
+      - "{{ if not .IsNightly }}flashbots/suave-geth:latest-amd64{{ end }}"
```
0x416e746f6e, line 100: **suggestion:**

```suggestion
      {{- if not .IsNightly }}
      - flashbots/suave-geth:latest-amd64
      {{ else }}
      - flashbots/suave-geth:latest-nightly-amd64
      {{- end }}
```

.goreleaser.yaml
```diff
@@ -106,8 +106,8 @@ dockers:
       - --platform=linux/arm64
     image_templates:
       - "flashbots/suave-geth:{{ .ShortCommit }}-arm64"
-      - "flashbots/suave-geth:{{ .Tag }}-arm64"
-      - "flashbots/suave-geth:latest-arm64"
+      - "flashbots/suave-geth:v{{ .Version }}-arm64"
+      - "{{ if not .IsNightly }}flashbots/suave-geth:latest-arm64{{ end }}"
```
0x416e746f6e, line 110: **suggestion:**

```suggestion
      {{- if not .IsNightly }}
      - flashbots/suave-geth:latest-arm64
      {{ else }}
      - flashbots/suave-geth:latest-nightly-arm64
      {{- end }}
```

.goreleaser.yaml
0x416e746f6e: yes, I think this whole block should look like:

```yaml
  {{- if not .IsNightly }}
  - name_template: "flashbots/suave-geth:latest"
    image_templates:
       - "flashbots/suave-geth:latest-amd64"
       - "flashbots/suave-geth:latest-arm64"
  {{ else }}
  - name_template: "flashbots/suave-geth:latest-nightly"
    image_templates:
       - "flashbots/suave-geth:latest-nightly-amd64"
       - "flashbots/suave-geth:latest-nightly-arm64"
  {{- end }}
```

.goreleaser.yaml
```diff
@@ -123,6 +123,12 @@ docker_manifests:
       - "flashbots/suave-geth:latest-amd64"
       - "flashbots/suave-geth:latest-arm64"
 
+nightly:
+  name_template: "{{ incpatch .Version }}-nightly"
+  tag_name: nightly
```
0x416e746f6e, line 128: **doubt:**

I am not very familiar with this part of goreleaser, but maybe it should be `tag_name: latest-nightly`.

.goreleaser.yaml
```diff
@@ -96,8 +96,12 @@ dockers:
       - --platform=linux/amd64
     image_templates:
       - "flashbots/suave-geth:{{ .ShortCommit }}-amd64"
-      - "flashbots/suave-geth:{{ .Tag }}-amd64"
-      - "flashbots/suave-geth:latest-amd64"
+      - "flashbots/suave-geth:v{{ .Version }}-amd64"
+      {{- if not .IsNightly }}
```
metachris, line 100: can you remove this inversion this for less cognitive load? i.e. instead of `{{- if not .IsNightly }}` do `{{- if .IsNightly }}`

metachris (approved):
overall lgtm. minor suggestion on removing the if inversion, but not a blocker

metachris (approved):

.goreleaser.yaml
0x416e746f6e: 😢 but I so love double negations 😢

PR #206, opened 2024-02-23
suave_structs.go: Marshal bytes to hexstring

jinmel:
- current: []byte field marshals to bass64 string
 - to be: BuildBlockArgs and DataRecord structs now use hexutil.Big and hexutil.Bytes for uint64 and []byte fields respectively

## 📚 References

---


suave/gen/main.go
```diff
@@ -243,6 +265,13 @@ type {{.Name}} struct {
 	{{range .Fields}}{{title .Name}} {{typ3 .Typ}}
 	{{end}}
 }
+
+type {{.Name | lower}}Marshaling struct {
```
ferranbt, line 269: Do you need to do this as part of the code generation?

Could you create another file `suave_structs_encoding.go` which is not code generated that does the encoding overriding?

core/types/suave_structs.go
```diff
@@ -22,6 +25,17 @@ type BuildBlockArgs struct {
 	FillPending    bool
 }
 
+type buildBlockArgsMarshaling struct {
```
ferranbt, line 28: The types that belong to suavex are handled now on suave-execution-geth. I already applied this fix there too https://github.com/flashbots/suave-execution-geth/pull/13

Makefile
```diff
@@ -45,6 +45,11 @@ devtools:
 
 suavedevtools:
 	go run ./suave/gen/main.go -write
+	cd core/types && \
```
ferranbt, line 48: Could you use // go:generate to do this? https://github.com/flashbots/suave-execution-geth/blob/feature/e2e-block-building-builderapi/suave/builder/api/api.go#L15

ferranbt (changes requested):

PR #209, opened 2024-02-26
Fixes gas estimation for ccrs

ferranbt:
## 📝 Summary

This PR fixes gas estimation for Confidential Compute requests. There were a few small issues here and there. Notes on the code as comments.

## 📚 References

---

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1027,6 +1027,8 @@ func DoCall(ctx context.Context, b Backend, args TransactionArgs, blockNrOrHash 
 			args.KettleAddress = &acc
 		}
 
+		// DoCall is also used in EstimateGas which does not have to include
+		// gas parameters or nonce. We need to set them to default values.
 		args.setDefaults(ctx, b)
 		tx := args.ToTransaction()
 
@@ -1048,6 +1050,9 @@ func DoCall(ctx context.Context, b Backend, args TransactionArgs, blockNrOrHash 
 			SkipAccountChecks: true,
 		}
 
+		// Run the MEVM but unlike with the send transaction endpoint, do not
+		// finalize the transactional store (third callback param). Otherwise,
+		// the updated kv entries will be committed to the store.
 		_, result, _, err := runMEVM(ctx, b, state, header, tx, msg, true)
 		if err != nil {
 			return nil, err
@@ -2017,7 +2022,10 @@ func runMEVM(ctx context.Context, b Backend, state *state.StateDB, header *types
 	}
 
 	if result.Failed() {
-		return nil, result, storeFinalize, nil
+		// If the execution fails we do not return an error but the transaction result itself
+		// since some callers of this function (i.e. estimate gas) differentiate between
+		// unrecoverable errors (i.e. nonce is incorrect) and failed executions.
+		return nil, result, nil, nil
 	}
 
 	if storageAccessTracer.hasStoredState {
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -47,7 +47,6 @@ import (
 	"github.com/ethereum/go-ethereum/params"
 	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/ethereum/go-ethereum/rpc"
-	suave "github.com/ethereum/go-ethereum/suave/core"
 	"github.com/tyler-smith/go-bip39"
 )
 
@@ -1028,6 +1027,7 @@ func DoCall(ctx context.Context, b Backend, args TransactionArgs, blockNrOrHash 
 			args.KettleAddress = &acc
 		}
 
+		args.setDefaults(ctx, b)
 		tx := args.ToTransaction()
 
 		state, header, err := b.StateAndHeaderByNumber(ctx, rpc.LatestBlockNumber)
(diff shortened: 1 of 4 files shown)


internal/ethapi/api.go
```diff
@@ -1028,6 +1027,7 @@ func DoCall(ctx context.Context, b Backend, args TransactionArgs, blockNrOrHash
 			args.KettleAddress = &acc
 		}
 
+		args.setDefaults(ctx, b)
```
ferranbt, line 1030: Estimate gas uses the `DoCall` function. We have to initialize the default values because the estimate gas is called without any gas limit or gas price.

internal/ethapi/api.go
```diff
@@ -1048,13 +1048,10 @@ func DoCall(ctx context.Context, b Backend, args TransactionArgs, blockNrOrHash
 			SkipAccountChecks: true,
 		}
 
-		_, result, finalize, err := runMEVM(ctx, b, state, header, tx, msg, true)
+		_, result, _, err := runMEVM(ctx, b, state, header, tx, msg, true)
```
ferranbt, line 1051: Estimate gas (and any other call) should not call finalize on the confidential store. Otherwise, the values are written for calls. This was not touched before on-purpose since we needed for the forge integration. Now it is not required anymore.

internal/ethapi/transaction_args.go
```diff
@@ -154,7 +154,7 @@ func (args *TransactionArgs) setFeeDefaults(ctx context.Context, b Backend) erro
 	}
 	// Now attempt to fill in default value depending on whether London is active or not.
 	head := b.CurrentHeader()
-	if b.ChainConfig().IsLondon(head.Number) {
+	if b.ChainConfig().IsLondon(head.Number) && !args.IsConfidential {
```
ferranbt, line 157: Do not use London pricing rules for confidential compute requests.

internal/ethapi/api.go
```diff
@@ -2017,7 +2017,7 @@ func runMEVM(ctx context.Context, b Backend, state *state.StateDB, header *types
 	}
 
 	if result.Failed() {
-		return nil, nil, nil, fmt.Errorf("%w: %s", result.Err, hexutil.Encode(result.Revert()))
+		return nil, result, storeFinalize, nil
```
ferranbt, line 2020: `result.Failed()` indicates an error during the execution but **it should not be treated** as a fatal error for the call. Functions that use `runMEVM` need to differentiate between fatal errors and execution errors.

dmarzzz (approved):

internal/ethapi/api.go
lthibault: I think this GitHub comment would make a great code-comment.

internal/ethapi/api.go
lthibault: Ditto.  Consider adding this as a code comment.

internal/ethapi/api.go
lthibault: Ditto.

internal/ethapi/transaction_args.go
lthibault: Ditto 🙃

suave/e2e/workflow_test.go
```diff
@@ -1045,65 +1050,27 @@ func TestRelayBlockSubmissionContract(t *testing.T) {
 	require.True(t, ok)
 }
 
-func TestE2E_ForgeIntegration(t *testing.T) {
-	// This end-to-end test ensures that the precompile lifecycle expected in Forge works
+func TestE2E_EstimateGas(t *testing.T) {
```
lthibault, line 1053: I recommend using `t.Parallel()` to cut down on CI time.

suave/e2e/workflow_test.go
```diff
@@ -1045,65 +1050,27 @@ func TestRelayBlockSubmissionContract(t *testing.T) {
...
+	fr.callPrecompile("newDataRecord", newDataRecordInput)
 
-	recordsRaw := doCall("fetchDataRecords", uint64(0), "default:v0:ethBundles")
-	var bids []types.DataRecord
-	require.NoError(t, mapstructure.Decode(recordsRaw[0], &bids))
-	require.Len(t, bids, 1)
-	require.Equal(t, bids[0].Id, record.Id)
+	// fetch the records and validate that it is empty since a call should not modify the store
+	vals := fr.callPrecompile("fetchDataRecords", []interface{}{uint64(0), "default:v0:ethBundles"})
+	require.Len(t, vals[0], 0)
```
lthibault, line 1066: Are we certain that `vals` always has a nonzero length?

suave/e2e/workflow_test.go
```diff
@@ -1045,65 +1050,27 @@ func TestRelayBlockSubmissionContract(t *testing.T) {
...
-	val := []byte{0x1, 0x2, 0x3}
-	doCall("confidentialStore", record.Id, "a", val)
+	// estimate the gas for the call should not modify the store either
+	gasLimit := fr.estimateGasPrecompile("newDataRecord", newDataRecordInput)
+	require.NotZero(t, gasLimit)
 
-	valRaw := doCall("confidentialRetrieve", record.Id, "a")
-	require.Equal(t, val, valRaw[0])
+	vals = fr.callPrecompile("fetchDataRecords", []interface{}{uint64(0), "default:v0:ethBundles"})
+	require.Len(t, vals[0], 0)
```
lthibault, line 1073: Same question about `len(vals)`.  Just want to make sure the test doesn't panic instead of failing cleanly.

lthibault (changes requested):
See inline for comments.  As a general note, the PR comments you wrote are very enlightening, and I think it would be beneficial to include them as code comments!

suave/e2e/workflow_test.go
ferranbt: Yup, enforced in `callPrecompile`.

lthibault (approved):
LGTM. Send it!

PR #210, opened 2024-02-26
add randomUint256 precompile

zeroXbrock:
## 📝 Summary

adds a new precompile `randomUint256` to generate a random number

## 📚 References

---


core/vm/contracts_suave.go
```diff
@@ -38,6 +40,16 @@ func (b *suaveRuntime) confidentialInputs() ([]byte, error) {
 	return b.contextGet("confidentialInputs")
 }
 
+func (b *suaveRuntime) randomUint256() (*big.Int, error) {
+	maxU256 := new(big.Int)
+	maxU256.SetString("FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF", 16)
```
lthibault, line 45: Why are we 1-ing out the U256?  It'll be zeroed out by default, so this feels unnecessary.

lthibault (approved):
Approving for expediency, but please see comment inline.

core/vm/contracts_suave.go
zeroXbrock: Can you explain how it gets zeroed out? I don't see where this happens. My understanding is that this defines the maximum value for the random number generator on the next line

lthibault: ☝️ Disregard.  This is setting the max random value, which is absolutely reasonable.  Resolving.

PR #211, opened 2024-02-27
Leak events from Confidential requests

ferranbt:
## 📝 Summary

This PR enables the leak of the events emitted during the offchain kettle computation in the callback return value. Then, a function in Solidity can parse those appended logs and emit them in Solidity. That part is covered in https://github.com/flashbots/suave-std/pull/56.

## 📚 References

---


suave/sol/standard_peekers/example.sol
```diff
@@ -44,6 +52,65 @@ contract ExampleEthCallSource {
...
+        for (uint256 i = 0; i < logs.length; i++) {
+            emit LogEvent(logs[i].addr);
+        }
+        */
+
+        // Call the function with decoded logs
+        _;
+    }
+
+    function emitLogCallback(uint256 num) public decodeLogs {
```
ferranbt, line 103: `decodeLogs` is the only modifier the callback needs to emit the logs from the confidential request.

internal/ethapi/api.go
```diff
@@ -2036,6 +2078,47 @@ func runMEVM(ctx context.Context, b Backend, state *state.StateDB, header *types
 		computeResult = result.ReturnData // Or should it be nil maybe in this case?
 	}
 
+	logs := state.GetLogs(common.Hash{}, 0, common.Hash{})
+
+	if len(logs) != 0 {
+		logs2 := []logTypeStruct{}
```
ferranbt, line 2084: Append the logs of the output to the result with some magic bytes as delimiter.

suave/sol/standard_peekers/example.sol
```diff
@@ -44,6 +52,65 @@ contract ExampleEthCallSource {
...
+        // revert("a");
+    }
+
+    event Example(uint256 num, uint256 indexed num2);
+
+    function emitLog() public payable returns (bytes memory) {
+        emit Example(1, 2);
```
ferranbt, line 110: Offchain computation emits an event that will be relayed on chain!

internal/ethapi/api.go
```diff
@@ -2044,6 +2044,24 @@ func runMEVM(ctx context.Context, b Backend, state *state.StateDB, header *types
...
+		}
+
+		logsEncoded, err := result.EncodeABI()
+		if err != nil {
+			return nil, nil, nil, err
+		}
+
+		// add some magic number and the result (TODO)
+		computeResult = append(computeResult, []byte{0xff}...)
+		computeResult = append(computeResult, logsEncoded...)
```
Ruteri, line 2061: I guess an interesting alternative would be to make the log coalescing explicit/configurable
As an example we could add a precompile that would return the coalesced logs, or have the logs live in the context and add a simple solidity wrapper for encoding the logs to the result of a function
This way we don't have to change the client code to change how logs are added and parsed

suave/sol/standard_peekers/example.sol
```diff
@@ -44,6 +44,44 @@ contract ExampleEthCallSource {
...
+
+        // Initialize memory for the data to decode
+        bytes memory dataToDecode = new bytes(dataLength);
+
+        // Copy the data to decode into the memory array
+        for (uint256 i = 0; i < dataLength; i++) {
+            dataToDecode[i] = inputData[magicSequenceIndex + i];
+        }
+
+        emit LogCallback(dataToDecode);
```
Ruteri, line 66: I wonder
The most obvious way to store the events would be encoded as an array of (event) bytes
This code points to that not being the case - rather we have a single encoded value (`ExecResult` which uses `tuple(tuple(address addr, bytes32[] topics, bytes data)[] logs)` underneath)
Could you simply encode that tuple as a struct and decode it here in Solidity?
What you would get is abi-encoded list of logs which you can then emit as-is using `log3` (or something like that I think?)

Also, what if instead you kept the encoded event instead of logs, and simply emitted the encoded logs? I think that would actually make the logs trivially parseable, whereas as is you have to use a roundabout way to decode the actual logs emitted (I don't think you show how to do that yet)

I'd look like so:
```solidity
bytes[] memory raw_logs = abi.decode(dataToDecode, (bytes[]));
for raw_log in raw_logs { emit raw_log; }
```
with a simple change in how you encode the execution result (or the log).

internal/ethapi/api.go
ferranbt: Something like?

```
function onchain(uint256 num) {
}

function offchain() {
   emit Event();

   return abi.encodeWithSelector(this.onchain.selector, 1, Suave.getLogs());
}
```

Well, you still need to create a new precompile and there is the work of teaching users to put that last argument. But I think making it implicit also leans more towards the side effect model. Also, the changes are not that big on the client.

internal/ethapi/api.go
```diff
@@ -2044,6 +2047,24 @@ func runMEVM(ctx context.Context, b Backend, state *state.StateDB, header *types
...
+			Logs: logs,
+		}
+
+		logsEncoded, err := result.EncodeABI()
+		if err != nil {
+			return nil, nil, nil, err
+		}
+
+		// add some magic number and the result (TODO)
+		computeResult = append(computeResult, magicBytes...)
```
Ruteri, line 2063: We should probably only do this if we actually have anything to encode right?

suave/e2e/workflow_test.go
```diff
@@ -1050,6 +1050,54 @@ func TestRelayBlockSubmissionContract(t *testing.T) {
...
+
+	t.Log("Testcases for execution result encoding")
+	t.Log(hex.EncodeToString(execResultBytes))
+
+	// check topics size for each log
+	require.Len(t, execResult.Logs[0].Topics, 0)
+	require.Len(t, execResult.Logs[1].Topics, 1)
+	require.Len(t, execResult.Logs[2].Topics, 2)
+	require.Len(t, execResult.Logs[3].Topics, 3)
+	require.Len(t, execResult.Logs[4].Topics, 4)
```
Ruteri, line 1093: It'd be great to also see how you can decode the specific log
Ideally we get the actual logs rather than the LogCallback containing data

suave/sol/standard_peekers/example.sol
```diff
@@ -44,6 +44,44 @@ contract ExampleEthCallSource {
 
         return abi.encodeWithSelector(this.emptyCallback.selector);
     }
+
+    event LogCallback(bytes data);
+
+    function emitLogCallback(uint256 num) public {
+        // From the msg.input, the 'confidential context' sequence
+        // starts at index 37 (4 signbature bytes + 32 bytes for uint256 + 3 bytes for the magic sequence)
+        uint256 magicSequenceIndex = 4 + 32 + 3;
```
Ruteri, line 53: Would you also want to make sure that msg.data[magicSequenceIndex] is equal to the magic sequence bytes?

suave/sol/standard_peekers/example.sol
```diff
@@ -44,6 +44,44 @@ contract ExampleEthCallSource {
 
         return abi.encodeWithSelector(this.emptyCallback.selector);
     }
+
+    event LogCallback(bytes data);
```
Ruteri, line 48: I'd suggest maybe a different name, since this doesn't actually cause anything to be executed, but is simply emitted
Maybe LogTrace? OffchainLogs?

suave/sol/standard_peekers/example.sol
Ruteri: I think actually emitting the logs rather than a single log would be interesting - those could then be parsed as one would parse any other logs, rather than having to go through a layer of indirection

using log<N> I think could be simple enough, since they only go until 4 right?

internal/ethapi/api.go
ferranbt: There is a check before that only appends if `logs!=0` which means that there is some data to encode here.

suave/e2e/workflow_test.go
ferranbt: The logs at that point are already decoded. It is done in https://github.com/flashbots/suave-geth/pull/211/files#diff-04dd0655f18243b3d4950eed741c8b50d9868ff52727bc3da5d9f990c46aa061R1081.

suave/sol/standard_peekers/example.sol
ferranbt: This is only for an e2e test to make sure that the encoded appended date (we have to find a new name for this) can be decoded later. `suave-std` which is the production side of things for the Solidity counterpart does more checks. https://github.com/flashbots/suave-std/pull/56/files#diff-c3373d1c6d1df488550173a8c39686af3c68cf6c525b7ca2446afa8dbcef95fbR36

suave/e2e/workflow_test.go
Ruteri: Oooh

PR #229, opened 2024-03-27
Do not error on state change for transient objects

ferranbt:
## 📝 Summary

Since we fail the CCR when a state change happens, this type of patterns is not possible:
```
contract Suapp {
  function example() public {
      Other o = new Other();
  }
}

contract Other {
   string jsonrpc;

   constructor(string _jsonrpc) {
       jsonrpc = _jsonrpc;
   }
}
```

We use this pattern a lot in `suave-std` and it is helpful as a way to describe transient objects with state.

This PR fixes the issue and only fails if the user tries to make state changes for the Suapp itself.

## 📚 References

---


dmarzzz (approved):
LSTM

PR #241, opened 2024-04-29
Make buildEthBlock multichain

ferranbt:
## 📝 Summary

This PR introduces a new precompile `buildEthBlockTo` with the same inputs as `buildEthBlock` but with an extra string argument that defines the url address (or service name) for the chain to use.

## 📚 References

---


dmarzzz (approved):
LGTM, we should add to the suave precompile spec and or have @andytudhope update docs 🙏🏻

zeroXbrock (approved):

PR #242, opened 2024-05-01
remove suaveup root password requirement

### LINKED ISSUE #238: suaveup installation without root password
### Describe the feature you would like

running the suaveup installation script puts the executable in `/usr/local/bin`, which requires root access. it would be preferable to install by default in a user specific location to avoid asking for the root password.

### Additional context

_No response_

fiiiu:
## 📝 Summary

changes suave-geth install location in suaveup to avoid requiring root password. closes #238 .

### DIFF
--- a/suave/suaveup/suaveup
+++ b/suave/suaveup/suaveup
@@ -56,20 +56,19 @@ else
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


fiiiu:
still need to add new location to `$PATH`.

ferranbt (approved):

PR #247, opened 2024-05-27
Sign CCRs as EIP-712 messages

ferranbt:
## 📝 Summary

This PR introduces CCRs that are signed with EIP-712 signatures. It introduces two changes:

- A new `signer/core/eip712` package with the `EIP-712` logic copied from `signer/core/apitypes` which due to some dependencies of the package `types` was not possible to use without a circular dependency. So, I copied to the new package only the core primitives.
- A new field `Envelope` in the `ConfidentialComputeRequest` object. If the field `Envelope` is not set, we use the [`eip-2718`](https://eips.ethereum.org/EIPS/eip-2718) rules to generate the signing hash. If the field `Envelope` is set, we generate the signing hash as an `EIP-712` envelope.

All the tests in e2e are using by default the new EIP-712 signing scheme.
 
## 📚 References

---

### DIFF
--- a/core/types/confidential.go
+++ b/core/types/confidential.go
@@ -18,7 +18,7 @@ type ConfidentialComputeRecord struct {
 	ConfidentialInputsHash common.Hash
 
 	// Envelope signals whether this CCR was signed using EIP-712
-	Envelope bool
+	IsEIP712 bool
 
 	ChainID *big.Int
 	V, R, S *big.Int
@@ -33,7 +33,7 @@ func (tx *ConfidentialComputeRecord) copy() TxData {
 		Gas:                    tx.Gas,
 		KettleAddress:          tx.KettleAddress,
 		ConfidentialInputsHash: tx.ConfidentialInputsHash,
-		Envelope:               tx.Envelope,
+		IsEIP712:               tx.IsEIP712,
 
 		Value:    new(big.Int),
 		GasPrice: new(big.Int),
--- a/core/types/confidential.go
+++ b/core/types/confidential.go
@@ -17,6 +17,9 @@ type ConfidentialComputeRecord struct {
 	KettleAddress          common.Address
 	ConfidentialInputsHash common.Hash
 
+	// Envelope signals whether this CCR was signed using EIP-712
+	Envelope bool
+
 	ChainID *big.Int
 	V, R, S *big.Int
 }
@@ -30,6 +33,7 @@ func (tx *ConfidentialComputeRecord) copy() TxData {
 		Gas:                    tx.Gas,
 		KettleAddress:          tx.KettleAddress,
 		ConfidentialInputsHash: tx.ConfidentialInputsHash,
+		Envelope:               tx.Envelope,
 
 		Value:    new(big.Int),
 		GasPrice: new(big.Int),
--- a/core/types/suave_eip712.go
+++ b/core/types/suave_eip712.go
@@ -2,7 +2,6 @@ package types
 
 import (
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/common/math"
 	"github.com/ethereum/go-ethereum/signer/core/eip712"
 )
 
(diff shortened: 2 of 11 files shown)


core/types/confidential.go
```diff
@@ -17,6 +17,9 @@ type ConfidentialComputeRecord struct {
 	KettleAddress          common.Address
 	ConfidentialInputsHash common.Hash
 
+	// Envelope signals whether this CCR was signed using EIP-712
+	Envelope bool
```
dmarzzz, line 21: nit: this feels not very descriptive

suave/cstore/engine.go
```diff
@@ -65,8 +65,8 @@ type CStoreEngine struct {
 	storage        ConfidentialStorageBackend
 	transportTopic StoreTransportTopic
 
-	daSigner    DASigner
-	chainSigner ChainSigner
+	daSigner DASigner
+	// chainSigner ChainSigner
```
dmarzzz, line 69: why do we get rid of chain signer?

suave/cstore/engine.go
```diff
@@ -262,10 +262,12 @@ func (e *CStoreEngine) Finalize(tx *types.Transaction, newRecords map[suave.Data
 		StoreUUID:   e.storeUUID,
 	}
 
-	if _, sigErr := e.chainSigner.Sender(tx); sigErr != nil {
-		log.Info("confidential engine: refusing to send writes based on unsigned transaction", "hash", tx.Hash().Hex(), "err", sigErr)
-		return suave.ErrUnsignedFinalize
-	}
+	/*
+		if _, sigErr := e.chainSigner.Sender(tx); sigErr != nil {
```
dmarzzz, line 266: delete?

suave/cstore/engine.go
```diff
@@ -317,10 +319,12 @@ func (e *CStoreEngine) NewMessage(message DAMessage) error {
 		log.Info("Confidential engine: message is spoofing our storeUUID, processing anyway", "message", message)
 	}
 
-	_, err = e.chainSigner.Sender(message.SourceTx)
-	if err != nil {
-		return fmt.Errorf("confidential engine: source tx for message is not signed properly: %w", err)
-	}
+	/*
+		_, err = e.chainSigner.Sender(message.SourceTx)
```
dmarzzz, line 323: delete?

suave/cstore/engine.go
```diff
@@ -367,11 +371,13 @@ func (e *CStoreEngine) NewMessage(message DAMessage) error {
 			return fmt.Errorf("confidential engine: caller %x not allowed on record %x", sw.Caller, sw.DataRecord.Id)
 		}
 
-		// TODO: move to types.Sender()
-		_, err = e.chainSigner.Sender(sw.DataRecord.CreationTx)
-		if err != nil {
-			return fmt.Errorf("confidential engine: creation tx for record id %x is not signed properly: %w", sw.DataRecord.Id, err)
-		}
+		/*
+			// TODO: move to types.Sender()
```
dmarzzz, line 375: delete?

core/types/confidential.go
ferranbt: Agree, what about `EIP712SigningSchema: bool`?

banr1: Since you two may not fully grasp the context, please let me know if my comments are off the mark.

This is a boolean, so I feel that using a format like `isXxx` would improve readability. What do you think? 
For example, how about simply using `isEIP712: bool` or something similar?

zeroXbrock: I like `IsEIP712` or `SignedWithEIP712`

core/types/suave_eip712.go
```diff
@@ -0,0 +1,52 @@
...
+	return eip712.TypedData{
+		Types: eip712.Types{
+			"EIP712Domain": []eip712.Type{
+				{Name: "name", Type: "string"},
+				{Name: "chainId", Type: "uint256"},
+			},
+			"ConfidentialRecord": []eip712.Type{
+				{Name: "nonce", Type: "uint64"},
+				{Name: "gasPrice", Type: "uint256"},
+				{Name: "gas", Type: "uint64"},
```
zeroXbrock, line 28: should this `uint64` be a `uint256`?

PR #249, opened 2024-05-30
Add derivePublicAddress precompile

pistomat:
## 📝 Summary

This PR adds a new precompile for deriving the public address from a SECP256 private key.
It aims to complement the [privateKeyGen](https://github.com/flashbots/suave-std/tree/main/src/suavelib/Suave.sol#L328) precompile. 

Right now the only way (as far as I know) to derive the address from a private key inside a confidential compute request is to sign a predetermined hash and derive the address using the `ecrecover` precompile, which is not very user friendly.

I extended the `TestE2E_Precompile_CryptoSECP256` test to also test this new precompile.

## 📚 References

- The `private-suapp-key-gen` example still has a `TODO` to return the address from the Suapp and validate the signature, this precompile enables it.
- The `crosschain-NFT-mint` uses the `vm.addr()` cheatcode from Forge to derive the address, but this is not available in actual Suapps.


ferranbt:
Hi @pistomat.

We are trying to limit the number of precompiles and if possible move as much logic to [suave-std](https://github.com/flashbots/suave-std) in Solidity as possible. AFAIK you can derive a public key from a private key in Solidity ([example](https://github.com/amiller/andromeda-sirrah-contracts/blob/main/src/crypto/secp256k1.sol#L67)). 

I will close this and open an issue in [suave-std](https://github.com/flashbots/suave-std/issues/87).

PR #275, opened 2024-08-08
secp256k1 encrypt/decrypt precompiles

zeroXbrock:
add secp256k1 en/decrypt precompiles

## 📝 Summary

- `secp256k1Encrypt`
- `secp256k1Decrypt`

## 📚 References

---


dmarzzz (changes requested):
test case plzz

suave/gen/suave_spec.yaml
```diff
@@ -445,3 +445,33 @@ functions:
         - name: value
           type: bytes
           description: "Randomly-generated bytes"
+  - name: secp256k1Encrypt
```
ferranbt, line 448: Can we create a generic encrypt/decrypt pair of precompiles and put both AES and this implementation there?

core/vm/contracts_suave.go
```diff
@@ -3,14 +3,19 @@ package vm
...
+	"crypto/sha256"
+	"encoding/binary"
 	"fmt"
 	"io"
 	"net/http"
 	"net/url"
 	"strings"
 	"time"
 
+	"github.com/decred/dcrd/dcrec/secp256k1/v4"
```
ferranbt, line 18: Could we do this with the crypto libraries already in Geth?

ferranbt (changes requested):

Issue #167, opened 2024-01-19
End Builder Session on New Block


dmarzzz:
# Rationale

Right now a session can last indefinitely but should be canceled when a new block comes in as that makes the state its simulating on invalid

# Implementation

Do you have ideas regarding the implementation of this feature?
Are you willing to implement this feature?

allnil:
Want to take a look

PR #281, opened 2024-10-06
End builder session on a New Block

7suyash7:
#167 

Implement a mechanism to end builder sessions when a new block arrives, to avoid simulations on invalid states.

- `SessionManager` now subscribes to `ChainHeadEvent`
- On receiving a new block event, all active sessions are terminated.
- Added checks to prevent new sessions from being created after `SessionManager` is closed.
- Added a `Terminate` method to `builder` for clean session ending.
- Added tests.

### DIFF
--- a/suave/builder/builder.go
+++ b/suave/builder/builder.go
@@ -1,6 +1,8 @@
 package builder
 
 import (
+	"context"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/state"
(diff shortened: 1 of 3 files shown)


suave/builder/builder.go
```diff
@@ -75,3 +82,9 @@ func (b *builder) AddTransaction(txn *types.Transaction) (*types.SimulateTransac
 
 	return result, nil
 }
+
+func (b *builder) Terminate() {
+	b.cancelFunc()
+	b.txns = nil
```
ferranbt, line 88: Why do you need to set this to nil?

ferranbt (approved):

suave/builder/session_manager.go
```diff
@@ -161,3 +183,57 @@ func (s *SessionManager) AddTransaction(sessionId string, tx *types.Transaction)
...
+
+		if timer, exists := s.sessionTimers[id]; exists {
+			timer.Stop()
+			delete(s.sessionTimers, id)
+		}
+
+		select {
+		case s.sem <- struct{}{}:
+		default:
+			panic("released more sessions than are open")
```
ferranbt, line 218: Can you error and not panic?

ferranbt (changes requested):

suave/builder/builder.go
7suyash7: Misunderstood how context works here, fixed it.

suave/builder/session_manager.go
7suyash7: Done

ferranbt (approved):

PR #285, opened 2024-11-12
fix: Iterate pendingWrites backward to get the last updated value

pistomat:
## 📝 Summary

Multiple `Store` operations in a single CCR append the values to `pendingWrites`, which makes it sorted by time ascending. When doing a `Retrieve` the list of `pendingWrites` was iterated from oldest to newest, which means only the first stored value for a given key was ever returned. This PR iterates over `pendingWrites` in reverse so the latest stored value is returned.

---


0x416e746f6e (approved):
reviewing upon request from Michał

looks good for my rather limited understanding of the problem and underlying code-base

